PR #177, opened 2019-02-08
Wireport

### LINKED ISSUE #168: WirePorts
Re-add "WirePorts".

Basically, they are very simple Components that are used to split wires and are rendered as a single port.

Example:
screen shot 2019-02-05 at 11 18 22 am

Try to write simple unit tests before implementing them to make sure they act appropriately.

Functionally, they act just like BUFGates.

For implementing them they'd be Component with exactly 1 input and output.

For rendering, you'd probably want to just put an if-statement before the rendering of the IOPorts in the ComponentRenderer.ts checking if the object is a WirePort and if it is then have a specific rendering process for them and 'return' early.
Or, potentially a better solution would be to set the origin/target of each input/output port to be at the origin of the local transform and the rendering of the port should just automatically happen but it could cause some rendering issues.

nprisament:
Adds wireports to the refactor

does not implement inserting wireports, deleting wireports, or reassigning wireports.

Fix #168

### DIFF
--- a/site/public/ts/controllers/MainDesignerController.ts
+++ b/site/public/ts/controllers/MainDesignerController.ts
@@ -258,7 +258,6 @@ export var MainDesignerController = (function() {
             s2.setPos(V(-200, -100));
             g1.setPos(V(0, 0));
             l1.setPos(V(200, 0));
-            wp.setPos(V(-100, 100));
 
             designer.addObjects([s1, s2, g1, l1]);
 
--- a/site/public/ts/controllers/MainDesignerController.ts
+++ b/site/public/ts/controllers/MainDesignerController.ts
@@ -30,7 +30,6 @@ import {ANDGate}  from "../models/ioobjects/gates/ANDGate";
 import {ORGate}  from "../models/ioobjects/gates/ORGate";
 import {XORGate}  from "../models/ioobjects/gates/XORGate";
 import {LED}      from "../models/ioobjects/outputs/LED";
-import {WirePort}      from "../models/ioobjects/WirePort";
 
 export var MainDesignerController = (function() {
     var designer: CircuitDesigner;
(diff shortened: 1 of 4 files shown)


tests/public/ts/models/ioobjects/WirePort.test.ts
```diff
@@ -0,0 +1,24 @@
+import "jest";
+
+import {CircuitDesigner} from "../../../../../site/public/ts/models/CircuitDesigner";
+import {WirePort}   from "../../../../../site/public/ts/models/ioobjects/WirePort";
+import {Switch}   from "../../../../../site/public/ts/models/ioobjects/inputs/Switch";
+import {LED}             from "../../../../../site/public/ts/models/ioobjects/outputs/LED";
+
+describe("Constant Low", () => {
```
LeonMontealegre: Describe should describe the component you're testing

site/public/ts/models/ioobjects/WirePort.ts
```diff
@@ -0,0 +1,32 @@
...
+
+	public getDisplayName() {
+		return "Port";
+	}
+
+	public getImageName() {
+		return "";
+	}
+
+    public getXMLName(): string {
```
LeonMontealegre: Spacing is weird here

site/public/ts/controllers/MainDesignerController.ts
```diff
@@ -30,6 +30,7 @@ import {ANDGate}  from "../models/ioobjects/gates/ANDGate";
 import {ORGate}  from "../models/ioobjects/gates/ORGate";
```
LeonMontealegre: Remove changes to MainDesignerController before pushing please 🙏

tests/public/ts/models/ioobjects/WirePort.test.ts
nprisament: fixed

site/public/ts/models/ioobjects/WirePort.ts
nprisament: fixed, for some reason, the tabs don't seem to be 4 spaces in this github viewer

site/public/ts/controllers/MainDesignerController.ts
nprisament: done

PR #178, opened 2019-02-08
[WIP] Actions

KevinMackenzie:
Added rotate and translate actions that form the backbone for the rotate and translate tools.  The action manager was slightly modified to that it doesn't implicitly execute the action on 'add'.  Rather, the actions should be in a post-executed state when added.  This improves usability for actions that are updated in real time like rotate and translate.


LeonMontealegre:
Let's hold off merging this into master for a bit until we get a full system working.

Rylander77:
nice

PR #181, opened 2019-02-08
Implemented Seven segment display and SSD renderer

jfhfbm:
Wrote SSD renderer in ComponentRenderer.ts. Still doesn't render segments...You said "I'll take care of it"


nprisament:
@jfhfbm you are going to have to fix these merge conflicts before this can be merged, it seems like the button things got on this branch for some reason.

PR #187, opened 2019-02-12
dry out the flipflop constructor

nprisament:


site/public/ts/models/ioobjects/flipflops/FlipFlop.ts
```diff
@@ -11,8 +11,8 @@ export abstract class FlipFlop extends Component {
 	protected state: boolean = false;
 	protected last_clock: boolean = false;
 
-    constructor(numInputs: number, numOutputs: number, size: Vector) {
-        super(new ClampedValue(numInputs), new ClampedValue(numOutputs), size);
+    constructor(numInputs: number) {
```
LeonMontealegre: This is unwanted behavior. Some flip-flops will be different sizes from one-another, @Schmaj is currently doing that with the addition of labels.

Removing numOutputs is okay but we need size to be specifiable

nprisament: In that branch, they are all being updated to 80x80, if they are all going to stay the same size (maybe two sizes if there is a labels on / labels off size) then shouldn't it be in the parent class?

nprisament: if not in this PR, I think that this should at least be done in general

LeonMontealegre: Just in general, the idea of each component being whatever size should be up to the individual classes - not the parent. The size is almost entirely a front-end thing and should be easily adjustable in case we ever want to change it in the future

LeonMontealegre:
This is failing TravisCI checks

nprisament:
fixed

Issue #182, opened 2019-02-08
Get LED light to light up


LeonMontealegre:

nprisament:
PR #193

LeonMontealegre:
Don't reference the PR from the issue, reference the issue from the PR by typing "Fix #ISSUENUMBER" and when the PR is merged it automatically closes the issue.
@nprisament

PR #193, opened 2019-02-22
Wrote LED class and wrote renderer for LED when it's lit

jfhfbm:
Modified ComponentRenderer.ts and created LED.ts


LeonMontealegre:
Fix #182

PR #196, opened 2019-02-23
Selection popup (title & position only)

baxtea:
Port count, color, etc popup modules coming soon

Also changes a handful of other things such as:
 - Project name and license now strictly follow `package.json` grammar
 - Fixed spelling error in `CircuitDesigner.propagate`
 - Fixes a bug in `SelectionTool` that allowed the translate tool to activate when there should not have been anything to drag
 - Adds a few new functions to `MainDesignerController`
 - Add a `clamp` function to `Vector`


site/public/ts/utils/selectionpopup/SelectionPopupModules.ts
```diff
@@ -0,0 +1,148 @@
...
+    // Called whenever an object is selected or deselected
+    // Also enables or disables the module depending on whether or not all the selected objects have the property this module needs
+    abstract pull(): void;
+    // Pushes changes to the current selection
+    // Called whenever the associated input element changes value
+    // As a consequence, this is only called when the module is enabled
+    abstract push(): void;
+}
+
+class Title extends Module {
```
LeonMontealegre: Please put each module in a different file

site/public/ts/utils/tools/SelectionTool.ts
```diff
@@ -122,24 +138,20 @@ export class SelectionTool extends Tool {
                 // Go through each object and see if it's within
                 //  the selection box
                 var objects = this.designer.getObjects();
+                let new_objs = false;
                 for (let obj of objects) {
                     // Check if object is in box
                     if (TransformContains(box, obj.getTransform())) {
                         // Add to selections if not already selected
-                        if (!this.selections.includes(obj))
+                        if (!this.selections.includes(obj)) {
```
LeonMontealegre: Did you mean to also set new_objs = true in this if-statement?

site/public/ts/utils/tools/SelectionTool.ts
```diff
@@ -238,6 +253,12 @@ export class SelectionTool extends Tool {
         return this.p2.copy();
     }
 
+    set onSelectionChanged(func: {(): void}) {
```
LeonMontealegre: Again, I'd rather just have an "addSelectionChangeListener" method or something rather than a setter

site/public/ts/utils/tools/SelectionTool.ts
```diff
@@ -238,6 +253,12 @@ export class SelectionTool extends Tool {
...
+        this.callbacks.push(func);
+    }
+    private selectionChanged() {
```
LeonMontealegre: selectionsChanged()*
Also let's keep private helper methods underneath the constructor my usual layout for classes are:

[variables]
[constructor]
[private methods]
[protected methods]
[public modifiers]
[public setters]
[public getters]
[abstract methods]

site/public/ts/controllers/MainDesignerController.ts
```diff
@@ -31,11 +31,13 @@ import {ANDGate}  from "../models/ioobjects/gates/ANDGate";
 import {ORGate}  from "../models/ioobjects/gates/ORGate";
 import {XORGate}  from "../models/ioobjects/gates/XORGate";
 import {LED}      from "../models/ioobjects/outputs/LED";
+import { SelectionPopupController } from "./SelectionPopupController";
```
LeonMontealegre: {SelectionPopupController}*

site/public/ts/utils/tools/SelectionTool.ts
baxtea: Yes, good catch

site/public/ts/controllers/MainDesignerController.ts
baxtea: With-spaces is the preferred syntax in the TS community. Still go ahead with the changes? If so, there are more instances of the same syntax in a few other files, not all from this PR. Should I change those too, or is it more about consistency per-file?

site/public/ts/controllers/MainDesignerController.ts
```diff
@@ -184,9 +193,18 @@ export var MainDesignerController = (function() {
         PlaceComponent: function(component: Component) {
             toolManager.placeComponent(component);
         },
+        GetSelections: function(): Array<IOObject> {
+            return toolManager.getSelectionTool().getSelections();
+        },
         GetDesigner: function(): CircuitDesigner {
             return designer;
         },
+        CanvasToScreen: function(p: Vector): Vector {
```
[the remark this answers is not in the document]
baxtea: In order to do this after making it a singleton, wouldn't I need to replace these with a `GetCamera` to use instead? Not sure if that's preferable... Alternatively, I could initialize the selection popup singleton inside of the main designer singleton's initializer, but that seems more confusing than keeping all of the `Init` calls in main. Suggestions?

LeonMontealegre:
@baxtea Alright, I made a few formatting changes and fixed a few small bugs with the Position Module.
Looks good!

Issue #201, opened 2019-02-28
Latches not drawing box


LeonMontealegre:

nprisament:
I think this has something to do with the svg for latches. I think that they haven't been reimplemented in the refactor

PR #206, opened 2019-03-01
fix some rendering stuff

nprisament:
#201


site/public/ts/models/ioobjects/latches/Latch.ts
```diff
@@ -30,6 +30,6 @@ export abstract class Latch extends Component {
 	}
 
 	public getImageName() {
-		return "latch.svg";
+		return "";
```
LeonMontealegre: if you’re just drawing a box the “image” should be blank

LeonMontealegre: whoops - mobile view of GitHub is messed up

PR #218, opened 2019-03-11
add labels to latches and clean up their rendering

nprisament:
I chose to make the latches 10 units smaller than the flipflops in both direction as to differentiate them from the FlipFlops of the same types


nprisament:
looking back over it, I like this change

PR #248, opened 2019-03-22
Added the ability to deselect using shift.

### LINKED ISSUE #217: Deselecting
When selecting a group of components (eg to form an IC) you are able to add more components to the selection, however the user is unable to deselect a component that is currently a part of the selection.

It would be useful to be able to deselect so as to not force the redoing of a large selection after a mistake. 

The desired behavior would be that if a currently selected component is clicked on while actively selecting (Shift key is held down) the component will be deselected (without deselecting the rest of the selection group).

tnguyen57:
I also added a library to tsconfig.js to stop my compiler from warning about the array's include function. The code still works fine without the library.

Fix #217

### DIFF
--- a/site/public/ts/controllers/MainDesignerController.ts
+++ b/site/public/ts/controllers/MainDesignerController.ts
@@ -55,8 +55,10 @@ export const MainDesignerController = (function() {
     }
 
     const onClick = function(button: number): void {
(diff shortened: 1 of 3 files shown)


LeonMontealegre:
You have to "npm install" again because we updated some packages (hence why the .includes" wasn't working)

PR #275, opened 2019-04-09
Wire ports snap and straighten neighboring wires

### LINKED ISSUE #243: Make straight wires work again
Also, have wires snap near points where they would be straight

baxtea:
Closes #243 

Also changes the functionality from the live site so dragging a set of components unstraightens all wires connecting them to unselected components.


baxtea:
After some experimentation, this was just due to using else if instead of if. Thoughts on moving the code to `WirePort.setPos`? I'm interested in the edge case of rotation: if snapping is moved to `setPos` instead of translate, then wires are unstraightened when rotated.

PR #280, opened 2019-04-12
Demux fix

### LINKED ISSUE #271: Demultiplexer does not work
When changing control signals, the outputs that were on stay on which is wrong

Schmaj:
Fixed the output count and refresh for the Demux.
Fix #271


Schmaj:
plz

PR #294, opened 2019-04-19
Nixed var...

jainr3:
Felt kinda obligated to nix var in my code, also I wanted to make a commit so I would be *officially* added to the contributors list! :)


nprisament:
woohoo! I might have made this a const but either way is eons better than var.

PR #295, opened 2019-04-25
Wire port deletion

### LINKED ISSUE #174: Snipping WirePorts
Wire Ports should be able to be cleanly deleted.

When they are deleted or cut, they should find the output port that the input wire is connected to and they should find the input port that the output wire is connected to and connect those two ports with a new wire.

rohanhpatel:
Added port deletion for exactly one port in wire connecting an Input IOObject and an Output IOObject

Fix #174


nprisament:
fix #174

PR #325, opened 2019-07-10
Replaced PHP backend with simple Go backend

KevinMackenzie:
In an effort to make the golang backend more easily reviewable, this is the first of several PRs merging changes from `golang-backend` (the branch).

This aims to replace the PHP page serving code with minimal go code.  To run this, you will need to install go, and clone the OpenCircuits repo in `$GOPATH$/github.com/OpenCircuits/OpenCircuits` (usually `$GOPATH$=<home>/go`).  Then, navigate to the `site/go` folder and run `go build`.  It will produce a binary, which you should run with `build/` as the working directory after running `npm run build`.

Since this simply serves the page, the goal is to make sure the interface is complete.

And yes, google insists that go should indent with tabs.


KevinMackenzie:
I'm looking into how to modify our docker configuration to support this (since it needs to be changed).

LeonMontealegre:
Fixing docker atm.
Added some more NPM scripts which will help with building.
To build docker just do npm run build:docker and then to run opencircuits just do npm start

PR #336, opened 2019-07-14
Added Google Auth and Stub for Auth Mock

KevinMackenzie:
Added authentication to the go backend that has a full implementation with the google sign-in api and a stub for an untrusted, username-only login for devs that do not want to setup API keys for google.  To enable no-auth sign in, use the `-no_auth` command-line flag

In order to test with google, you'll need to create a web app with the sign-in API (https://developers.google.com/identity/sign-in/web/sign-in click). Choose "New Project", give it a name,  give it a product name, and choose "Web browser".  For the "Authorized JavaScript Origins", enter: `http://localhost:8080`.  Click "Create..." and this will give you a client ID and secret.  Save these somewhere in the following format and DO NOT commit them to the repo.
```
{
    "id": "<client-id>.apps.googleusercontent.com",
    "secret": "<secret>",
    "redirectURL": "http://127.0.0.1:8080/auth/google"
}
```

To enable google sign-in, you must add the command-line flag `-google_auth=<path-to-file>`.

If this is working properly, clicking "sign in with google" in the "sign-in" pop up and you should be prompted without a window that asks you to sign in with your credentials.

Things that still should be done before merging:
- [x] Make the UI better for logging in.  Currently it recycles the visual style from the download-type dropdown (help much appreciated)
- [x] Fully implement the "no_auth" login method so devs don't need google API keys to use the back-end
- [x] Add command-line flags for enabling different auth methods 
- [x] Add real google token verification (its just a placeholder now)


LeonMontealegre:
TODO: Add to wiki:
go get github.com/gin-gonic/contrib/sessions
go get golang.org/x/oauth2
go get golang.org/x/oauth2/google

KevinMackenzie:
hmm, I didn't have to do anything extra to get the dependencies to work.  a single `go get` might automatically grab everything it needs but I can't test that without a fresh go environment

LeonMontealegre:
Apparently, all the "comments" on exported functions have to be in the form of
`// FUNCTIONNAME description`

site/public/ts/controllers/LoginController.ts
```diff
@@ -0,0 +1,151 @@
...
+
+                // Load 'auth2' from GAPI and then initialize w/ meta-data
+                await new Promise<void>((resolve) => gapi.load('auth2', resolve));
+                await gapi.auth2.init(new class implements ClientConfig {
+                    // Written like this to make ESLint happy
+                    public client_id?: string = googleAuthMeta.getAttribute("content");
+                }).then(async (_) => {}); // Have to explicitly call .then
+
+            }
+            else if (noAuthMeta) {
```
KevinMackenzie: is there any reason in particular this is an `else if` instead of an `if?  It would be useful to use `if` so we can make sure it works with multiple auth methods at the same time.

PR #406, opened 2019-10-01
Fix #403 by adding a clickable GitHub logo to go to the GitHub repo! :)

### LINKED ISSUE #403: Add link to GitHub repository
Add a link to the GitHub straight from the page. Could look something like this:

untitled

or

untitled (1)

maybellekusumoto:
See #403 

Now the header shows as follows:
image

With the GitHub logo clickable(!!) :)

Fix #403


Rylander77:
Fix #304

site/public/scss/_header.scss
```diff
@@ -88,6 +88,25 @@ header {
...
 }
 
+.header__center__github {
+    position: absolute;
+    top: 10px;
+    bottom: 0;
+    left: 50%;
+
+    width: 40px;
+    height: 40px;
```
LeonMontealegre: isn't this automatically set? or no?
If not, let's calculate it using $header-height - 2*$header-padding

site/public/scss/_header.scss
```diff
@@ -88,6 +88,25 @@ header {
     }
 }
 
+.header__center__github {
```
LeonMontealegre: I had .header__center__logo as it's own thing since it was the only header__center thing but now that we have this too, we should wrap both of them around a .header__center class that share some attributes like position, left: 50%, height, color, and the @media thing.

site/templates/header.tmpl
```diff
@@ -14,6 +14,9 @@
 
     <div class="header__center">
         <img id="logo" class="header__center__logo" src="img/icons/logo.svg" height="100%" alt="OpenCircuits logo"/>
+        <a href="https://github.com/OpenCircuits/OpenCircuits/" target="_blank">
+          <img id="github" class="header__center__github" src="img/icons/github.svg" height="100%" alt="GitHub logo"/>
```
LeonMontealegre: Let's also indent this by 4 spaces not 2 to be consistent

PR #410, opened 2019-10-04
LED Culling

rsdexter:
Fixed the issue of LEDs not culling properly. Overrode the getMinPos() and getMaxPos() methods to increase the cullBox size only while the LED is on. Also overrode the activate() method to update the cull box as the LED is turned on/off.


app/digital/ts/models/ioobjects/outputs/LED.ts
```diff
@@ -18,6 +21,57 @@ export class LED extends Component {
...
+            p => p.getWorldTargetPos().add(IO_PORT_RADIUS+IO_PORT_BORDER_WIDTH)
+        );
+
+        return Vector.max(max, ...corners, ...ports);
+    }
+
+    // @Override
+    public activate(signal: boolean, i: number = 0): void {
+        this.onTransformChange();
+        this.getCullBox();
```
LeonMontealegre: Only need to call this.onTransformChange and cullBox should update automatically.

app/digital/ts/models/ioobjects/outputs/LED.ts
```diff
@@ -18,6 +21,57 @@ export class LED extends Component {
...
+        const ports = this.getPorts().map(
+            p => p.getWorldTargetPos().add(IO_PORT_RADIUS+IO_PORT_BORDER_WIDTH)
+        );
+
+        return Vector.max(max, ...corners, ...ports);
+    }
+
+    // @Override
+    public activate(signal: boolean, i: number = 0): void {
+        this.onTransformChange();
```
LeonMontealegre: Don't forget to call super.activate

app/digital/ts/models/ioobjects/outputs/LED.ts
```diff
@@ -19,6 +25,57 @@ export class LED extends DigitalComponent {
         this.inputs.first.setTargetPos(V(0, 2*DEFAULT_SIZE));
     }
 
+    // @Override
+    public getMinPos(): Vector {
+        const min = V(Infinity);
```
LeonMontealegre: We need to use super.getMinPos() and super.getMaxPos for getMaxPos, so that the "find minimum pos for ports" doesn't have to be done in here.
See: https://github.com/OpenCircuits/OpenCircuits/pull/412/files

Issue #311, opened 2019-06-14
Clock doesn't copy/save it's set frequency


LeonMontealegre:

LeonMontealegre:
Same with LED + Color

PR #411, opened 2019-10-04
Fixed clocks not saving their frequency

dwisdom751:
added a specific save and load function for saving clocks, so their frequency is saved.  Before clocks only used the general save function for components which didn't take into account the additional frequency variable that clocks have.  Closing #311


app/digital/ts/models/ioobjects/inputs/Clock.ts
```diff
@@ -2,6 +2,8 @@ import {V} from "Vector";
 import {ClampedValue} from "math/ClampedValue";
 
 import {Component} from "../Component";
+import {XMLNode}      from "core/utils/io/xml/XMLNode";
```
LeonMontealegre: 1 Space between {XMLNode} and from

app/digital/ts/models/ioobjects/inputs/Clock.ts
```diff
@@ -47,4 +49,14 @@ export class Clock extends Component {
     public getImageName(): string {
         return (this.isOn ? "clockOn.svg" : "clock.svg");
     }
+
+    public save(node: XMLNode): void {
```
LeonMontealegre: Also update the copy method so that the Clock copies correctly

PR #413, opened 2019-10-04
Firefox compatibility and friends

baxtea:
Tinting and select-highlighting components works properly on Firefox, LED elements are tinted when selected like other components, and added a simple method of `preventDefault()`ing certain bound keys which are known to be disruptive (e.g. we don't want backspace to ever navigate to the previous page instead of deleting the selection).


app/digital/ts/rendering/ioobjects/ComponentRenderer.ts
```diff
@@ -105,17 +106,23 @@ export const ComponentRenderer = (() => {
...
             if (object instanceof LED) {
-                if (object.isOn())
-                    renderer.image(Images.GetImage(object.getOnImageName()), V(), size.scale(3), object.getColor());
+                // draw the LED object
+                renderer.image(Images.GetImage(imgName), V(), size, object.getColor());
+
+                // draw the LED glow
+                if (object.isOn()) {
+                    const glow_img = Images.GetImage(object.getOnImageName());
+                    const LED_GLOW_SIZE = V(ROTATION_CIRCLE_RADIUS, ROTATION_CIRCLE_RADIUS).scale(2);
```
LeonMontealegre: You can just do `V(ROTATION_CIRCLE_RADIUS)` and it will set it for both x,y

app/digital/ts/rendering/ioobjects/ComponentRenderer.ts
```diff
@@ -105,17 +106,23 @@ export const ComponentRenderer = (() => {
...
+                    renderer.image(glow_img, V(), LED_GLOW_SIZE, object.getColor());
+                }
+
+                // tint green on top if selected
+                if (tint !== undefined)
+                    renderer.overlayTint(Images.GetImage(imgName), V(), size, tint);
             }
+            else if (imgName)
+                renderer.image(Images.GetImage(imgName), V(), size, tint);
```
LeonMontealegre: Braces {} for else if since the if has braces

PR #414, opened 2019-10-04
Add tooltips to header icons

### LINKED ISSUE #348: Hovering over website icons should provide tooltips
The icons below when hovered over do not provide tooltips describing the action. The icons are pretty good but for accessibility tooltips should exist.

image

shailpatels:
Icons on the upper right header now have icons for the buttons that are image only.
Added for dropdown menu aswell

foo2

closes #348


shailpatels:
This should close the issue it references so I think it's done. 
I can do more accessibility improvements in this branch or start another if there are more issues.

LeonMontealegre:
@shailpatels 
That issue was made a long time ago when those were basically the only things in the header.
So it might be good to add more for the login window and sidenav to the left and even the item nav for placing objects.

shailpatels:
Sounds good! I'll mark this as WIP and update the issue.
I think tooltips for placing objects would be good too - that might be a separate issue/pr since its a bit more complex then updating the HTML templates

PR #418, opened 2019-10-17
Added the flags for IP address and Port

### LINKED ISSUE #399: Add runtime flags for webserver settings (i.e. Port #)
The go webserver is run with the default options and a non-conflicting port number (8080). We want to be running this on 80/443 on production.  The IP address is set to `0.0.0.0`, which might not be what we need for a deployed instance.

We need to 
1. Determine which of the default / current settings should be runtime-configurable
2. Add run time flags for setting these variables instead of hard-coding them.

maybellekusumoto:
Fixes #399 

Added the flags `--ip_address` and `--port`, but kept the default IP Address of `0.0.0.0` and port of `8080`!! :)

### DIFF
--- a/site/go/main.go
+++ b/site/go/main.go
@@ -65,6 +65,5 @@ func main() {
 	authManager.RegisterHandlers(router)
 	api.RegisterRoutes(router, authManager, exampleCsif, userCsif)
 
-	// TODO: add flags for this
 	router.Run(*ipAddressConfig + ":" + *portConfig)
 }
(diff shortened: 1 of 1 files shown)


Rylander77:
The default IP address can be changed to `127.0.0.1` if need be, then can be changed using the flags.

LeonMontealegre:
@KevinMackenzie

site/go/main.go
```diff
@@ -64,5 +66,5 @@ func main() {
 	api.RegisterRoutes(router, authManager, exampleCsif, userCsif)
 
 	// TODO: add flags for this
```
KevinMackenzie: Remove the `TODO` tag now that the feature is implemented :)

PR #422, opened 2019-10-18
Fixed WirePort Copying

### LINKED ISSUE #375: Copying a single WirePort works
Shouldn't be able to copy+paste a single WirePort

rosan2:
Fixed the problem where the digital node would be copied and there would just be a random node floating about, not connected to any wires.

Closes #375


site/public/ts/digital/controllers/DigitalCopyController.ts
```diff
@@ -25,13 +27,32 @@ export class DigitalCopyController extends CopyController {
...
         designer.addGroup(CopyGroup(objs));
 
+        const objs2 = designer.getObjects();
+        console.log(objs2);
+        for (const obj of objs2) {
+            if (isNode(obj)) {
+                console.log(obj);
+                console.log(obj.getConnections().length);
+                console.log(designer.getWires().length);
+                if ((obj.getConnections().length) < 2) {
```
LeonMontealegre: remove extra parenthesis (obj.getConnections().length) => obj.getConnections().length

site/public/ts/digital/controllers/DigitalCopyController.ts
```diff
@@ -25,13 +27,32 @@ export class DigitalCopyController extends CopyController {
...
+                    designer.removeObject(obj);
+                    if (designer.getWires().length == 1) {
```
LeonMontealegre: this is designer.getWires(), but don't we want to check obj.getConnections() still?
Cause designer.getWires().length == 1 is only true if there's 1 wire in the entire circuit

PR #427, opened 2019-10-21
Improve Docker Caching and set default port to 80

Rylander77:
The best caching for Docker is actually minimizing the number of layers, as seen [here](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers). As a result, we'll minimize stages to copying, installing dependencies, and building the components of the project.

Uses the new `port` flag in the command to serve on default HTTP port 80, allowing for normal HTTP connection to the IP of the Docker image on whatever network(s) the docker image is on.


Dockerfile
```diff
@@ -3,18 +3,22 @@
...
 FROM golang:1.13-alpine AS builder
 
-# Install git.
+# Install git and gcc
 # Git is required for fetching the dependencies.
-RUN apk update 
-RUN apk add git
-RUN apk add build-base
+# GCC is required for sqlite3
+RUN apk add --no-cache git build-base
```
LeonMontealegre: Why no cache?

Rylander77: Usage for `--no-cache` is seen [here](https://github.com/gliderlabs/docker-alpine/blob/master/docs/usage.md#disabling-cache).

PR #430, opened 2019-10-24
Adds "Delete Circuit" Support

### LINKED ISSUE #396: Add "Delete Circuit" support
There's no way to delete circuits yet.  Should be pretty obvious why we want it.  We should have a button on the UI, like in the side-bar where your circuits are listed, that lets you delete it, the a confirmation window, then it will be removed from the back-end permanently.  

Maybe in the future we hold onto the deleted circuits for a period of time so they can undo it, but if we have a confirmation on the front-end that should be fine to start

KevinMackenzie:
Closes #396 

Currently WIP.  The back-end support is added in general and the storage interfaces in master are supported.  Note that the in-memory storage interface does not delete the element from the array, but this is ok since this interface should only really be used in its full capacity for testing anyway.

Whats Left:
 - [x] Add front-end support (i.e. trash can icon in side-bar with confirmation window)


KevinMackenzie:
The behavior for delete is as followed:
- Press "delete" icon
- User is prompted with option to delete or cancel
- If the user clicks "ok", it will delete the circuit from the server and reload the list
- If the currently open circuit is deleted, it is deleted from the server, but the data is preserved locally

site/public/ts/shared/controllers/RemoteController.ts
```diff
@@ -96,5 +97,16 @@ export const RemoteController = (() => {
...
+        DeleteUserCircuit(metadata: CircuitMetadata, callback: (result: boolean) => Promise<void> | void = Promise.resolve): void {
+            Chain(async (data: RemoteData) => {
+                const result = await DeleteUserCircuit(data.authState, metadata.getId());
+                callback(result);
+                if (data.metadata !== undefined && data.metadata.getId() === metadata.getId()) {
+                    return {
+                        metadata: metadata.buildOn().withId("").build()
+                    }
+                }
+            });
```
LeonMontealegre: I believe deletion should require an auth state, so set the `requireAuth` parameter true when calling chain and it'll take care of the behavior of someone trying to delete without an auth state (which probably isn't something that's going to happen but better safe than sorry)

site/public/ts/shared/controllers/SideNavController.ts
```diff
@@ -81,7 +81,19 @@ export class SideNavController {
...
+                preview.onClick(
+                    () => RemoteController.LoadUserCircuit(d, (c) => this.loadCircuit(c)), 
+                    () => {
+                        if (confirm("Are you sure you want to delete circuit \"" + d.getName() + "\"?")) {
+                            RemoteController.DeleteUserCircuit(d, (succ) => {
+                                if (!succ) {
+                                    alert("Failed to delete circuit!");
+                                } else {
+                                    this.updateUserCircuits()
+                                }
```
LeonMontealegre: I'd prefer a good ol'
```
if (!succ) {
    alert("Failed to delete circuit!");
    return;
}
this.updateUserCircuits();

File: site/public/ts/shared/controllers/SideNavController.ts [hunk 3]
```diff
@@ -81,7 +81,19 @@ export class SideNavController {
...
+                            });
+                        }
```
LeonMontealegre: Let's also just put this whole thing in a method or something cause this is an unholy amount of nesting

LeonMontealegre:
Also please fix Travis-CI linting errors, you can lint locally with `npm run lint`

PR #435, opened 2019-10-29
Fixed Clock's Frequency bugs

dwisdom751:
I changed the clocks so they have their frequencies when saving and loading to and from outside files.  I also fixed some formatting from my previous change.  I also fixed copy so when clocks are copied the frequency of the new clock will be the same as the old clock.


dwisdom751:
I tried to push changes to clock.ts but when I tried travis CI failed.  However, the test that failed was the resistor test that failed but I did not make any changes to the resistor.  I'm not sure what to do.  The clock test passed.

Issue #398, opened 2019-09-20
Less aggressive cache-busting scheme for non-dev environments


KevinMackenzie:
The current method used for cache-busting files is to load them from the disk every time they are requested.  This helps development because any changes made to the ts when its built will be available when the page is refreshed.  This is not a very good plan for deployed instances since we do want to cache these files.

The debug-mode cache-busting can be the current implementation, but there should be a command-line switch for "is production" that lets us switch behavior like this for different subsystems.

KevinMackenzie:
We can concatenate the "last modified" time of the file to the end of the file name when serving static resources.

There are two flavors of this options as I see it:
 - Query file metadata every page load so we could hot-patch the static files without restarting the server
 - Query file metadata once on startup so we never have to touch the disc on page load

PR #436, opened 2019-11-01
Added another static resource caching scheme

KevinMackenzie:
Related to #398 

Previously, the file was always read from the disc on page load, which is fine for debugging, but keeping the data in memory is much faster.  The currently used cache scheme replicates the behavior of the previous caching scheme, but there is a new option that can be switched on for production builds in the future.


LeonMontealegre:
@KevinMackenzie is this ready to merge?

KevinMackenzie:
Probably

PR #440, opened 2019-11-02
Thumbnail renders in .circuit files

### LINKED ISSUE #423: Thumbnail for User-Circuits generation

baxtea:
Not used anywhere in the UI, but thumbnails are rendered as 90% quality base64-encoded 256x256 PNGs and stored in the XML tag <thumbnail> when downloaded. Rather confident this will work for any circuit (so long as the components used have properly-implemented cullboxes), but have only tested the examples. Copying circuits to clipboard uses the same `WriteCircuit` function, but thumbnails are not embedded in that case.

Example 3's thumbnail

Closes #423.


site/public/ts/shared/utils/Exporter.ts
```diff
@@ -0,0 +1,100 @@
...
+    const vw = THUMBNAIL_SIZE/window.innerWidth;
+    const vh = THUMBNAIL_SIZE/window.innerHeight;
+    const view = new DigitalCircuitView(canvas, vw, vh);
+
+    // Center and zoom the camera so everything fits with no distortion
+    const center = V((xmin + xmax)/2, (ymin + ymax)/2);
+    const camera = view.getCamera();
+    camera.setPos(center);
+    // Zoom out 2.5% more than we need, just for a little padding
+    const zoom = Math.max((xmax - xmin)/THUMBNAIL_SIZE, (ymax - ymin)/THUMBNAIL_SIZE) * 1.025;
```
LeonMontealegre: Let's make that another constant for the zoom padding

site/public/ts/shared/utils/Exporter.ts
```diff
@@ -0,0 +1,100 @@
...
+    camera.zoomBy(zoom);
+
+    // Create background the same color as the site to composite the image on top of
+    // NOTE: can't just fill the first canvas before the render since the first thing the renderer does it clear it
+    const composite = document.createElement("canvas");
+    composite.width = THUMBNAIL_SIZE;
+    composite.height = THUMBNAIL_SIZE;
+    const context = composite.getContext("2d");
+    context.fillStyle = window.getComputedStyle(document.body, null).getPropertyValue("background-color");
+    context.fillRect(0, 0, THUMBNAIL_SIZE, THUMBNAIL_SIZE);
```
LeonMontealegre: This seems really bad. Can we not just set the CSS background attribute on the first canvas to be the gray? I feel like that would work a lot better.

site/public/ts/shared/utils/Exporter.ts
```diff
@@ -0,0 +1,100 @@
...
+    context.drawImage(canvas, 0, 0, THUMBNAIL_SIZE, THUMBNAIL_SIZE);
+    return composite;
+}
+
+export function WriteCircuit(designer: DigitalCircuitDesigner, name: string, thumbnail: boolean = false): string {
+    const writer = new XMLWriter(designer.getXMLName());
+    writer.setVersion("1.1");
+    writer.setName(name);
+    if (thumbnail) {
+        // Render to a small temporary canvas (which gets garbage-collected once done)
```
LeonMontealegre: `document.createElement` creates the canvas, do we have to `document.removeElement` or something? I'd assume it really doesn't get garbage collected until removed from the DOM strictly. We should also make sure this is true because canvas's use a ton of memory.

site/public/ts/shared/utils/Exporter.ts
```diff
@@ -0,0 +1,100 @@
...
+        writer.setThumbnail(thumbnail);
+    } else {
+        writer.setThumbnail("data:,");
+    }
+
+    designer.save(writer.getContentsNode());
+
+    return writer.serialize();
+}
+export function SaveFile(designer: DigitalCircuitDesigner, projectName: string): void {
```
LeonMontealegre: Should this also have the `thumbnail: boolean = false` parameter?

site/public/ts/shared/views/CircuitView.ts
```diff
@@ -15,18 +15,16 @@ export abstract class CircuitView {
...
         this.canvas = canvas;
         this.renderer = new Renderer(this.canvas, vw, vh);
         this.camera = new Camera(this.canvas.width, this.canvas.height);
 
         this.resize();
     }
 
-    public render(designer: CircuitDesigner, selections: Selectable[], toolManager: ToolManager): void {
+    // toolManager changed to be optional because the exporter doesn't want to create a dummy tool manager
+    public render(designer: CircuitDesigner, selections: Selectable[], toolManager?: ToolManager): void {
```
LeonMontealegre: You passed in `undefined` when you called it above, so you should remove the `undefined` parameter passing since it's optional.

site/public/ts/shared/utils/Exporter.ts
baxtea: Up to you. Probably makes the most sense to add it, yeah.

site/public/ts/shared/utils/Exporter.ts
baxtea: I was also concerned about this, but eventually assumed it must be garbage collected since the only removal method I found at the time, `canvas.parentNode.removeChild` was crashing. On a second look through, there's another way to do it, just with `canvas.remove()`. Tested, it works.

site/public/ts/shared/utils/Exporter.ts
baxtea: That's because it is really bad lol. But neither `background` nor `backgroundColor` work to make it gray. If you have any other ideas, I'm all ears. Could just throw in a flag somewhere to tell the view not to clear before rendering, but that seems a little dirty in its own right.

site/public/ts/shared/utils/Exporter.ts
```diff
@@ -0,0 +1,100 @@
...
+    let ymax = -Infinity;
+    for (let c of designer.getObjects()) {
+        const cullbox = c.getCullBox();
+        const max = cullbox.getTopRight();
+        const min = cullbox.getBottomLeft();
+        xmin = Math.min(min.x, xmin);
+        ymin = Math.min(min.y, ymin);
+        xmax = Math.max(max.x, xmax);
+        ymax = Math.max(max.y, ymax);
+    }
```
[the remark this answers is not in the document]
baxtea: So there is not currently a way to get an axis-aligned bounding box for a component?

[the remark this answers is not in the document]
LeonMontealegre: There is, but the process of actually determining if an axis-aligned box is in the screen versus just a normal box is extremely expensive (since it's done every frame if zooming or panning) and kind of defeats the purpose of cull-boxes as an optimization (since cull-boxes only change when the component moves/changes)

site/public/ts/shared/utils/Exporter.ts
LeonMontealegre: Did you do .style.background?

LeonMontealegre: Wait, I just realized that the normal "download as png" doesn't have a background color. So yep I guess this is the only way

LeonMontealegre: We could also just have the thumbnail component display with a gray background in the actual display on the SideNav. I kind of like this idea more actually

baxtea: Agreed, last idea is better. Back to saving with a transparent background.

PR #443, opened 2019-11-02
Fix saving example circuits

### LINKED ISSUE #391: Saving example circuits breaks
Attempting to save an example circuit will try to save via `api/circuits/<circuit-id>` like a user's circuit even though the circuit ids for example circuits are only valid for `api/example/<circuit-id>`.  Additionally, example circuits are read-only.

We should decide how we want to handle this, but now the `RemoteController` becomes unresponsive after the backend returns a `404 not found`.  This should probably be a front-end change.

shailpatels:
Clicking save on an example issue would previously cause the RemoteController to send out an error. Changed the load Example Circuit method to pass back default metadata instead of one with example circuit specific IDs which do not exist for the user. 

fixes  #391


LeonMontealegre:
@shailpatels 
Can we check to see if saving the loaded example circuit keeps the name? It should keep the name but I have a feeling it might be "Untitled Circuit"

shailpatels:
@LeonMontealegre 
I think its saving correctly
image

Tested by logging in by noauth
selecting example circuit
changed name and hit save

PR #454, opened 2019-11-19
GCP Prod Setup

LeonMontealegre:
Fixed a few small backend issues for prod and added some package.json scripts for easier prod development.

@KevinMackenzie 
I had to temporarily hack in a way for some flags to get set for prod through env variables since you can't pass flags (from what I was able to research).
Definitely should fix it to be a better way before merging
```
if os.Getenv("DATASTORE_PROJECT_ID") != "" {
	*googleAuthConfig = "credentials.json"
	*userCsifConfig = "gcp_datastore"
}
```


KevinMackenzie:
I think that `GOOGLE_CLOUD_PROJECT` is probably a better environment variable indicating if the app is running in the production environment, but it isn't any better a solution.

We could do all configuration through environment variables since that is in the `app.yaml`.  It seems weird that we can't control the parameters of the application.

PR #513, opened 2020-03-18
Fit screen

sandorzm:
Added function `GetCameraFit` in `ComponentUtils` that returns a position and zoom to fit a camera to an array of `CullableObject` with adjustable padding. `ThumbnailGenerator`, and the newly added F key shortcut in `SelectionTool`, now use this function.

`SelectionTool` calls the function on the selections, if any; otherwise, it calls on _all_ `CullableObject`s. It also uses the new `MoveCameraAction` to allow the user to undo fit-to-screen actions.

### DIFF
--- a/app/core/ts/actions/camera/MoveCameraAction.ts
+++ b/app/core/ts/actions/camera/MoveCameraAction.ts
@@ -21,16 +21,6 @@ export class MoveCameraAction implements Action {
         this.finalZoom = finalZoom;
     }
 
-    // Static method conveniently creates a MoveCameraAction after the
-    // camera is moved (the constructor is for before the camera is moved)
-    public static postMoveCameraAction(camera: Camera, initialPos: Vector, initialZoom: number): MoveCameraAction {
-        const action = new MoveCameraAction(camera, camera.getPos(), camera.getZoom());
-        action.initialPos = initialPos;
-        action.initialZoom = initialZoom;
-
-        return action;
-    }
-
     public execute(): Action {
         this.camera.setPos(this.finalPos);
         this.camera.setZoom(this.finalZoom);
--- a/app/core/ts/actions/camera/MoveCameraAction.ts
+++ b/app/core/ts/actions/camera/MoveCameraAction.ts
@@ -1,0 +1,48 @@
+import {Vector} from "Vector";
+import {Camera} from "math/Camera";
+import {Action} from "core/actions/Action";
+
+export class MoveCameraAction implements Action {
+    protected camera: Camera;
+
+    protected initialPos: Vector;
+    protected finalPos: Vector;
+
+    protected initialZoom: number;
+    protected finalZoom: number;
+
+    public constructor(camera: Camera, finalPos: Vector, finalZoom: number) {
+        this.camera = camera;
+
+        this.initialPos = camera.getPos();
+        this.finalPos = finalPos;
+
+        this.initialZoom = camera.getZoom();
+        this.finalZoom = finalZoom;
+    }
+
+    // Static method conveniently creates a MoveCameraAction after the
+    // camera is moved (the constructor is for before the camera is moved)
+    public static postMoveCameraAction(camera: Camera, initialPos: Vector, initialZoom: number): MoveCameraAction {
+        const action = new MoveCameraAction(camera, camera.getPos(), camera.getZoom());
+        action.initialPos = initialPos;
+        action.initialZoom = initialZoom;
+
+        return action;
+    }
+
+    public execute(): Action {
+        this.camera.setPos(this.finalPos);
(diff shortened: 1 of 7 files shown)


app/core/ts/actions/camera/MoveCameraAction.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    // Static method conveniently creates a MoveCameraAction after the
+    // camera is moved (the constructor is for before the camera is moved)
+    public static postMoveCameraAction(camera: Camera, initialPos: Vector, initialZoom: number): MoveCameraAction {
+        const action = new MoveCameraAction(camera, camera.getPos(), camera.getZoom());
+        action.initialPos = initialPos;
+        action.initialZoom = initialZoom;
+
+        return action;
+    }
```
LeonMontealegre: This shouldn't be necessary, so let's delete and see comments below

app/core/ts/tools/SelectionTool.ts
```diff
@@ -209,6 +213,21 @@ export class SelectionTool extends DefaultTool {
...
+            const initialPos = this.camera.getPos();
+            const initialZoom = this.camera.getZoom();
+
+            // Fit to selections, if any; otherwise, fit all CullableObjects
+            const objs = this.selections.size === 0
+                ? (this.designer.getObjects() as CullableObject[]).concat(this.designer.getWires())
+                : this.getSelections().filter(o => o instanceof CullableObject) as CullableObject[];
+
+            FitCamera(this.camera, objs, FIT_PADDING_RATIO);
+            this.action.add(MoveCameraAction.postMoveCameraAction(this.camera, initialPos, initialZoom));
```
LeonMontealegre: Instead of having a `postMoveCameraAction`, we have a few options:
1. `FitCamera` returns a `MoveCameraAction` to the desired position and zoom
2. `FitCamera` returns a tuple of the final position and zoom, {pos, zoom} and then we create the regular action

I'll leave you to decide this, but I personally like the latter.

app/core/ts/utils/ComponentUtils.ts
```diff
@@ -307,6 +310,22 @@ export function CircuitBoundingBox(all: CullableObject[]): BoundingBox {
...
+
+    const screenSize = camera.getCenter().scale(2); // Bottom right corner of screen
+    const worldSize = camera.getWorldPos(screenSize).sub(camera.getWorldPos(V(0,0))); // World size of camera view
+
+    // Determine which bbox dimension will limit zoom level
+    const xRatio = bbox.getWidth()/worldSize.x;
+    const yRatio = bbox.getHeight()/worldSize.y;
+    camera.zoomBy(Math.max(xRatio, yRatio) * padding);
+}
+
```
LeonMontealegre: See above comment.

For 2) see [this](https://www.typescriptlang.org/docs/handbook/basic-types.html#tuple)

site/public/ts/shared/utils/ThumbnailGenerator.ts
```diff
@@ -24,21 +25,7 @@ export class ThumbnailGenerator {
...
-        //  with extra padding on sides
-        const center = min.add(max).scale(0.5);
-        const relativeSize = max.sub(min).scale(1/this.size);
-        const zoom = Math.max(relativeSize.x, relativeSize.y) * THUMBNAIL_ZOOM_PADDING_RATIO;
-
-        // Move camera
-        const camera = this.view.getCamera();
-        camera.setPos(center);
-        camera.setZoom(zoom);
+        FitCamera(this.view.getCamera(), all, THUMBNAIL_ZOOM_PADDING_RATIO);
```
LeonMontealegre: For 1) we'd also have to change this to use the action, which could easily be done by just changing this to

`FitCamera(this.view.getCamera(), all, THUMBNAIL_ZOOM_PADDING_RATIO).execute();`

app/core/ts/tools/SelectionTool.ts
sandorzm: I see how 1 would work, but `FitCamera` already kind of returns the final position and zoom in that it modifies the original camera object to have the final position and zoom.

Another thing I could do is make a copy of the camera before `FitCamera` and then used the regular `MoveCameraAction` constructor using that copy and the final position and zoom. (This might be best done by adding a `Camera.copy` method.)

LeonMontealegre: I don't think copying the entire camera is necessary.

I like 2) because it turns FitCamera from a function that changes the camera to one that just calculates some numbers which is usually better since it's easier to test. And if you go with 2) we should rename the function GetBestFit or something

sandorzm: Ok, I'll do 2

PR #519, opened 2020-04-07
Santized user input

Robert-Hammond:
User input for Output Count, Input Count, and Clock Frequency can no longer make components disappear, become unfathomably large, attempt to make NaN or non-integer number of ports, or have some number of ports beyond the max or below the min.


site/public/ts/digital/controllers/selectionpopup/OutputCountPopupModule.ts
```diff
@@ -51,11 +52,15 @@ export class OutputCountPopupModule extends SelectionPopupModule {
...
-        const countAsNumber = this.count.valueAsNumber;
-
-        this.circuitController.addAction(new GroupAction(
-            selections.map(o => new OutputPortChangeAction(o, countAsNumber))
-        ).execute());
+        let countAsNumber = this.count.valueAsNumber;
+        if (countAsNumber % 1 === 0) // cannot be NaN
+        {
+            countAsNumber = Clamp(countAsNumber, parseInt(this.count.min), parseInt(this.count.max));
+            this.count.value = countAsNumber.toString();
```
LeonMontealegre, line 59: This is being repeated so many times, we should make a shared method somewhere
My recommendation:

Make a sub-class of `SelectionPopupModule` called `NumberInputPopupModule` that the 3 modules will extend instead and it will have a
`    protected count: HTMLInputElement;` instance variable and maybe you can reduce the repetition of code in `OutputCountPopupModule`, `InputCountPopupModule`, and `ClockFrequencyPopupModule`, and for sure have a `getInput()` method which returns the sanitized and clamped input

site/public/ts/digital/controllers/selectionpopup/OutputCountPopupModule.ts
```diff
@@ -10,6 +10,7 @@ import {MainDesignerController} from "../../../shared/controllers/MainDesignerCo
 import {Encoder} from "digital/models/ioobjects/other/Encoder";
 
 import {SelectionPopupModule} from "../../../shared/selectionpopup/SelectionPopupModule";
+import {Clamp} from "math/MathUtils";
```
LeonMontealegre, line 13: Math imports at top

site/public/ts/digital/controllers/selectionpopup/InputCountPopupModule.ts
```diff
@@ -14,6 +14,7 @@ import {Decoder} from "digital/models/ioobjects/other/Decoder";
 import {Mux} from "digital/models/ioobjects/other/Mux";
 
 import {SelectionPopupModule} from "../../../shared/selectionpopup/SelectionPopupModule";
+import {Clamp} from "math/MathUtils";
```
LeonMontealegre, line 17: math imports at top

PR #530, opened 2020-06-26
Wire selection

### LINKED ISSUE #520: Wire's can't be clicked when they're in a "straight" mode
**Describe the bug**
Wire's keep their non-straight click-box while straight which makes it not seem clickable when straightened out.

**To Reproduce**
1. Go to 'opencircuits.io'
2. Create a Switch and LED
3. Connect them with a Wire
4. Move one of the components so that the wire snaps to a straightened mode
5. Try to click on the wire

**Expected behavior**
Wire should calculate it's click-able area differently when straightened

**Screenshots**
N/A

 - OS: MacOS
 - Browser: Chrome

**Additional context**
N/A

rsdexter:
Closes #520 . Wire's updateCurve function changes c1 and c2 of the Bezier curve to match p1 and p2 only while the Wire's 'straight' member variable is true, creating a straight Bezier curve that can be properly selected.


app/core/ts/models/Wire.ts
```diff
@@ -50,13 +50,25 @@ export abstract class Wire extends CullableObject {
...
             const dir = this.p1.getWorldDir();
             this.shape.setP1(pos);
-            this.shape.setC1(dir.scale(DEFAULT_SIZE).add(pos));
+            // To create a straight bezier curve, c1 must be at the same position as p1
+            if(this.straight) {
+                this.shape.setC1(pos);
+            }
+            else {
+                this.shape.setC1(dir.scale(DEFAULT_SIZE).add(pos));
+            }
```
LeonMontealegre, line 59: Fix formatting please:
space between `if` and parenthesis,
no brackets needed since it's a one-liner

app/core/ts/models/Wire.ts
```diff
@@ -50,13 +50,25 @@ export abstract class Wire extends CullableObject {
...
             const dir = this.p2.getWorldDir();
             this.shape.setP2(pos);
-            this.shape.setC2(dir.scale(DEFAULT_SIZE).add(pos));
+            // To create a straight bezier curve, c2 must be at the same position as p2
+            if(this.straight) {
+                this.shape.setC2(pos);
+            }
+            else {
+                this.shape.setC2(dir.scale(DEFAULT_SIZE).add(pos));
+            }
```
LeonMontealegre, line 71: Same as above

PR #547, opened 2020-07-09
Cookiebanner

SethLaurenceau:
Added banner under the header to comply with GDPR regulations, declining use of cookies redirects to google


site/public/ts/shared/controllers/HeaderController.ts
```diff
@@ -128,6 +128,14 @@ export class HeaderController {
 
             main.setLocked(!main.isLocked());
         });
+
+        $(".cookie__banner__accept").click(() => {
+            const banner = document.getElementById("cookie-banner");
+            banner.classList.add("invisible");
```
LeonMontealegre, line 134: Should just do `$("#cookie-banner").addClass("invisible")` since already using JQuery

site/public/scss/_header.scss
```diff
@@ -315,6 +314,43 @@ header {
     }
 }
 
+.cookie__banner {
```
LeonMontealegre, line 317: It'd be better to have `left: 33%` and `right: 33%` so that it's guaranteed to be centered instead of `width: 33%`.

Also, let's have it at the bottom of the page since that's generally where cookie banners lie, so `bottom: 0` and remove `top: xxx`.

I think some more padding on the top and bottom would be nice as well

LeonMontealegre (approved):

PR #548, opened 2020-07-31
shifted canvas down, too simple to take a week

### LINKED ISSUE #515: Banner partially blocks camera view
**Describe the bug**
The top banner of the site is considered part of the camera view, so when a tall circuit is fitted to the screen, part of the top of the view is blocked by the banner. This results in apparently much less padding above the circuit than below, or even some components being clipped off, which looks bad.

**To Reproduce**
1. Run branch "master" locally
2. Create a circuit that's very tall vertically
3. Press F to fit to screen
4. Observe uneven padding

**Expected behavior**
Padding should be between *bottom of banner* and top of circuit, but it's between *top of browser tab* and top of circuit and thus is blocked by the banner.

**Screenshots**
padding-bug

 - OS: Windows 10 Education 1909, build 18363.720
 - Browser: Firefox 74.0 (64-bit)

SethLaurenceau:
strangely enough, shifting down by the full header height overshot and produced the opposite bug, hence the half header shift

closes #515


SethLaurenceau:
was that just there when you loaded the site?
what browser are you using?
OS?

It doesn't appear on my end (Firefox 76 on linuxmint)
so any clues to tease it out( on my windows box would if need be) would be swell

Robert-Hammond:
Yeah it just looked like that when I started it.
Chrome on Windows, tried it again just now and the same issue for Chrome, Firefox, and Edge.
Analog works perfectly, it only looks like that on digital.

SethLaurenceau:
I am unable to test this on a windows machine, docker requires a feature that my windows install refuses to provide

Robert-Hammond:
I'll see if I can fix it

Robert-Hammond:
I noticed the "position" box in the right hand side graphic shows a top position of 21.435:
image
When I uncheck `position: absolute;` in the `.canvas` field, the positioning border goes away:
image
The same thing happens when I uncheck the `position: fixed;` in the `.header` field, which is probably a better option.
For some reason my browsers don't use the most recently built version of stylesheet.css, so it's up to you to make changes.
I would try changing the _header.scss file first, changing `position: fixed;` to `absolute` (also add `top: 0px;` in this case) or just commenting it out.
Best of luck!

LeonMontealegre (approved):

PR #587, opened 2021-02-23
Fixed issue w/ split wire tool

### LINKED ISSUE #585: Status Indicator for Saving
Should have a little "loading circle" when saving a circuit, and a red "X" symbol when it failed to load. Should be right next to the save button.

LeonMontealegre:
Closes #585 

Fixed the issue where the Split Wire Tool didn't work at all


LeonMontealegre (commented):
I think you might need to remove the `.execute()` on the SplitWireTool as well

Issue #588, opened 2021-02-24
Clicking and dragging on wires leads to crash


Xystem4:
At any point trying to drag on a wire and create a node will lead to an immediate crash. 

Dragging and rotating a wire does not cause any issues, only splitting does.

PR #589, opened 2021-02-26
Split wire fix

Xystem4:
Closes issue #588

Relocates execution of SplitWireAction to inside the creation of the group, rather than after the creation (as the later actions in the group could not properly be created before the prior actions were executed).

Also added proper creation of the vector holding the initial position of the node when performing the split wire action (in SplitWireTool). (also renamed said vector from initalPosition to initialPosition).

Remaining issue from this is that now wires can be split, but after doing so existing nodes mid-wire cannot be reselected and moved (aside from typing in number coordinates). Dragging on an existing node just creates a new one. However I believe this issue is arising from a separate area of the code.

### DIFF
--- a/src/app/core/tools/SplitWireTool.ts
+++ b/src/app/core/tools/SplitWireTool.ts
@@ -56,6 +56,9 @@ export const SplitWireTool: Tool = (() => {
             action.add(CreateSplitWireAction(designer, wire, port));
 
             info.currentlyPressedObject = port;
+
+            // Set initial position
+            initialPosition = new Vector(camera.getWorldPos(input.getMousePos()));
         },
         onDeactivate({}: Event, {history}: CircuitInfo): void {
             history.add(action.add(new TranslateAction([port], [initialPosition], [port.getPos()])));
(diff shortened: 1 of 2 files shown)


src/app/core/tools/SplitWireTool.ts
```diff
@@ -15,7 +15,7 @@ import {Node, Wire} from "core/models";
 
 
 export const SplitWireTool: Tool = (() => {
-    let initalPosition: Vector;
+    let initialPosition: Vector;
```
LeonMontealegre, line 18: Whoops! Good catch!

src/app/core/tools/SplitWireTool.ts
```diff
@@ -53,12 +53,15 @@ export const SplitWireTool: Tool = (() => {
             // Set wireport as selection and being pressed
             action.add(CreateDeselectAllAction(selections).execute());
             action.add(new SelectAction(selections, port).execute());
-            action.add(CreateSplitWireAction(designer, wire, port).execute());
+            action.add(CreateSplitWireAction(designer, wire, port));
 
             info.currentlyPressedObject = port;
+
+            // Set initial position
+            initialPosition = new Vector(camera.getWorldPos(input.getMousePos()));
```
LeonMontealegre, line 61: You can just have `initialPosition = camera.getWorldPos(input.getMousePos());`

LeonMontealegre (changes requested):

LeonMontealegre (approved):

Issue #521, opened 2020-04-10
Double clicking wires sometime's gets caught on SelectionPopup


LeonMontealegre:
**Describe the bug**
When double-clicking a wire in certain positions, it causes the SelectionPopup to show up in front of the mouse and then prevents the 2nd click from registering

**To Reproduce**
1. Go to 'opencircuits.io'
2. Create a Switch and an LED right of it
3. Connect them with a wire
4. Try to double click on the right part of the wire

**Expected behavior**
Wire should be double-clickable without SelectionPopup interfering.
This could be done in multiple ways.
My immediate idea is to but a small buffer on the SelectionPopup showing up, maybe specifically just for wires. Maybe also somehow penetrate through the popup and register the second click event if done quick enough.

**Screenshots**
<img width="587" alt="Screen Shot 2020-04-10 at 1 24 14 PM" src="https://user-images.githubusercontent.com/29997057/79009890-a4187600-7b2e-11ea-9886-341ceecce1ab.png">
Red area is where the mouse clicked

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser: Chrome

**Additional context**
N/A

PR #593, opened 2021-02-26
Wire double click fix

JumpingKangaroo:
Closes issue #521 

Allows users to click through a SelectionPopup for 500ms after it is shown. After that point, it becomes clickable. This will let the user double click a wire to select all connected wires, even if the wires are in a configuration where normally the box popup would block the wire.

I tested this out with the fix to wire selection that was just merged to master and was able to double click through it for half a second.

### DIFF
--- a/src/site/shared/containers/SelectionPopup/index.tsx
+++ b/src/site/shared/containers/SelectionPopup/index.tsx
@@ -1,6 +1,6 @@
 import {useLayoutEffect, useState} from "react";
 
-import {HEADER_HEIGHT} from "shared/utils/Constants";
+import {HEADER_HEIGHT, DOUBLE_CLICK_DURATION} from "shared/utils/Constants";
 
 import {V} from "Vector";
 import {Camera} from "math/Camera";
@@ -70,7 +70,7 @@ export function SelectionPopup({info, modules}: Props) {
                     { 
                         return { ...prevState, clickthrough }
                     }),
-                500);
+                DOUBLE_CLICK_DURATION);
             }
 
             lastPos = pos;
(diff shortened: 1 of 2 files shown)


Xystem4:
Seems to properly take care of the issue. Works regardless of what portion of the popup box is obstructing the mouse (you can control this by zooming in until the wire takes the entire screen).

One note that is that the double clicking works even if something else is obstructing the wire (a node, a gate, another wire, etc.), which may or may not be the desired functionality. I believe this is not a result of these changes however, but has been present since the double-click functionality was added.

Xystem4 (approved):

src/site/shared/containers/SelectionPopup/index.tsx
```diff
@@ -55,7 +61,17 @@ export function SelectionPopup({info, modules}: Props) {
             if (pos === lastPos && visible === lastVisible)
                 return;
 
-            setState({ pos, visible });
+            setState({ pos, visible, clickthrough });
+
+            // Set up a callback that will make the box clickable if it's currently not clickable
+            if (clickthrough === true) {
+                clickthrough = false;
+                setTimeout(() => setState((prevState) =>
```
LeonMontealegre, line 69: Let's write this as:
```
setTimeout(() => 
    setState((prevState) => ({ ...prevState, clickThrough })),
    DOUBLE_CLICK_DURATION
);
```

LeonMontealegre (changes requested):

JumpingKangaroo:
@LeonMontealegre Whenever you get a chance, made the changes you recommended above!

LeonMontealegre (approved):

PR #606, opened 2021-03-06
Menu now closes if you click the same menu icon again

### LINKED ISSUE #611: Clicking the icon of an open dropdown does not close the dropdown
When you click the icon of an open dropdown (such as the download dropdown) it does not close the dropdown. This is currently being addressed in #606

TGCrystal:
Previously when a header menu was open, say the download menu, clicking the download icon again wouldn't close it. Now it does, just like it does on opencircuits.io
I figured out how to fix it before I even got around to creating an issue for it so there is no issue to link to
Fixes #611

### DIFF
--- a/src/site/shared/state/Header/reducers.ts
+++ b/src/site/shared/state/Header/reducers.ts
@@ -13,8 +13,7 @@ export function headerReducer(state = initialState, action: AllSharedActions): H
         case OPEN_HEADER_MENU_ID:
             return {
                 ...state,
-                // Close the menu by clicking on its icon while it is open
-                curMenu: ((state.curMenu === action.menu) ? ("none") : (action.menu))
+                curMenu: action.menu
             };
         case OPEN_HEADER_POPUP_ID:
             return {
--- a/src/site/shared/state/Header/reducers.ts
+++ b/src/site/shared/state/Header/reducers.ts
@@ -13,7 +13,8 @@ export function headerReducer(state = initialState, action: AllSharedActions): H
         case OPEN_HEADER_MENU_ID:
             return {
                 ...state,
-                curMenu: action.menu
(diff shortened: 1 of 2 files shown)


src/site/shared/state/Header/reducers.ts
```diff
@@ -13,7 +13,8 @@ export function headerReducer(state = initialState, action: AllSharedActions): H
         case OPEN_HEADER_MENU_ID:
             return {
                 ...state,
-                curMenu: action.menu
+                // Close the menu by clicking on its icon while it is open
+                curMenu: ((state.curMenu === action.menu) ? ("none") : (action.menu))
```
LeonMontealegre, line 17: This totally works, but in general, the reducers and their state should be as "dumb" as possible. They pretty much directly take in actions and cause a logical change in state. And so when I, as a programmer, call "OpenHeaderMenu", I expect the menu to be open and not toggle. I think the better change would be in `src/site/shared/containers/Header/Right/Dropdown.tsx` and change it so that when you click it and it's already open, it closes it instead.

LeonMontealegre (changes requested):

src/site/shared/state/Header/reducers.ts
TGCrystal: The onclick behavior is defined for each separate dropdown rather than unified in ```Dropdown.tsx```. Should each separate dropdown be modified so that it calls a new function such as ```ToggleHeaderMenu```, or should just the name of the dropdown be passed into ```Dropdown.tsx``` so that Dropdown can decide whether to call OpenHeaderMenu or CloseHeaderMenus? (or is there some other way that I'm missing?)

LeonMontealegre (approved):

Issue #596, opened 2021-02-27
Buttons stay pressed


Xystem4:
When pressing a button it never switches back to the off state, as far as I can tell.

4 buttons pressed!
![Screenshot 2021-02-27 010148](https://user-images.githubusercontent.com/65842664/109377147-6e3fc580-7897-11eb-9763-4b2f7e7a6221.png)

PR #620, opened 2021-03-23
Button fix

Xystem4:
Closes issue #596. It fixes the Interaction Tool onEvent method to call release rather than press on a mouseup event. It also adds an onActivate method to DefaultTools and InteractionTools, so that after a drag action is completed the mouseup action can be registered. (this onActivate method is called from the ToolManager).


src/app/core/tools/DefaultTool.ts
```diff
@@ -10,6 +10,10 @@ export class DefaultTool {
         this.handlers = handlers;
     }
 
+    public onActivate(event: Event, info: CircuitInfo): boolean{
```
LeonMontealegre, line 13: this can return `void` and also there should be a space b/t `void` and `{`

src/app/core/tools/ToolManager.ts
```diff
@@ -26,6 +26,7 @@ export class ToolManager {
                 // Deactivate the tool
                 this.currentTool.onDeactivate(event, info);
                 this.currentTool = undefined;
+                this.defaultTool.onActivate(event,info);
```
LeonMontealegre, line 29: space after the comma

LeonMontealegre (changes requested):

LeonMontealegre (approved):

Issue #597, opened 2021-02-27
Mobile bugs v2


TGCrystal:
There are a few mobile bugs not addressed by #492 
These bugs are present on both Safari and Chrome on an iPhone 6s Plus unless otherwise noted.

1. Dragging items from the itemnav to the canvas does not work at all
2. The download and tours/resources dropdown do not close when tapping on the canvas
3. In portrait mode, the download icon is pushed below the header (see screenshot below)

![IMG_4462](https://user-images.githubusercontent.com/29614966/109377697-012e2f00-789b-11eb-89c9-f7feb689685c.PNG)

Issue #298, opened 2019-05-31
Touch Pen Support


LeonMontealegre:
By default, it selects.
When pressing button, it'll pan

PR #627, opened 2021-03-30
Mobile drag fix

LeonMontealegre:
This is like the 10th iteration of drag-n-drop fixes, mostly cause they have terrible support within browsers...
Finally gave up and wrote a solution "from 1st principles", tested on Chrome/Firefox w/ virtual mobile and on Safari on iPhone.

We desperately need tests for this functionality once we have front-end tests.

Fixes part 1 of #597


TGCrystal:
I think this partially addresses #298 
I tested it in Chrome and Firefox on my touchscreen laptop with a stylus pen and with my finger and both of them allowed components to be placed and dragged around after they were placed. Chrome did get confused and think I was trying to gesture to go back a page though. Still, I couldn't drag anything from the itemnav on the live or nightly site, but I could on this branch.
I did find that regardless of platform, the preview when dragging from the itemnav to the canvas was not present. This was the case regardless of browser, operating system, or desktop/mobile.

PR #628, opened 2021-03-31
Update bug_report.md

Xystem4:
I added four missing colons to the bug report template that were making me disproportionately upset


LeonMontealegre:
We need to submit an issue to Github itself! Cause I stole their default bug report template!

LeonMontealegre (approved):
I love it

Issue #468, opened 2019-12-07
ICs don't propagate signals immediately when connected


LeonMontealegre:
Create IC with Switch + LED (switch on)

Then delete the wire between Switch and LED and replace it with IC hooked up, the LED won't turn on.

Probably because the LED was on when the IC was created and so an "ON" signal didn't change the signal so it didn't propagate. When hooking up connections on the IC we need to turn the ports on/off when created based on what the LEDs are.

TGCrystal:
Likely due to the same underlying issue as #613

PR #630, opened 2021-04-03
Logical Boolean Expression to Circuit Generator

### LINKED ISSUE #566: Create IC from given logical expression
Creating large and complex IC's is time consuming. Also, there is no quick feature to convert a logical expression to a circuit.

A quick way to create an IC would be to enter a logical expression such as "!(p|q)&r". This would create an IC with inputs "p", "q", and "r" and output "output". The name of the IC would be the expression used to create it. The IC would store a circuit representation of the given expression. The creation of the IC would be done through a prompt that shows up when dragging an "Expression" option from the "Other" section of the component UI onto the screen.

**Describe alternatives you've considered**
The benefit of using an IC as opposed to a new component is that it would benefit from all the features that an IC has or will have later. For example, if issue #528 is implemented, this could be used to generate a truth table from a logical expression. 

**TODO**
 Create tests, there are lots of corner cases because of parenthesis to consider
Create the expression parser
Use the expression parser to generate an IC
Add something to the "other" section of the component list

TGCrystal:
I'm currently creating this draft pull request to help organize my thoughts on what I have done and have left to do to close #566 

So far, you can enter an expression, and the circuit it represents is placed and "spread out."
The place to go to do this is under a new utilities dropdown, shown here (it should only show up in the digital site):
image
Currently, this is what the input UI looks like: 
image
It certainly leaves a lot to be desired. Here is what that same circuit looks like when Generate is pressed:
image
This works, but it also leaves a lot to be desired. Also, I currently have the coordinates of the top right component set to 0,0, so if the screen isn't moved at all this entire thing starts in the middle and expands to the bottom right. The distances between components probably needs some tweaking which is an easy fix, but it still leaves many of the components obstructing wires. That may not be an easy fix.

Overall, here is a list of changes still yet to be made, roughly grouped into categories of priority:

High Priority:
- [x] Fix placement of circuit so that the coordinates start at the top left of whatever the user currently sees
- [x] ~~Refactor the Popup system to use a DigitalPopup (quick start popup and keyboard shortcuts popup both have their state string defined in the Header state in the core despite having their actual code in the digital side, so this is just following that paradigm)~~ issue for later
- [x] Improve UI of entry screen (copy style for save button to the generate button, do something to make the input field look better, style the error message)
- [x] Do something for invalid input error handling (ideally displaying the error message somewhere)
- [x] Detect an error for something like "a b"
- [x] ~~Any other problems with the code that would prevent the pull request from being accepted~~ redundant

Medium Priority:
- [x] Generate as IC option
- [x] Select input component
- [x] Allow for different variations of logical symbols in input (|| and + for or, && and * for and, etc.)
- [x] Placeholder to show user intended input format
- [x] Clear equation after closing the entry popup
- [x] Rename placed inputs to match given expression
- [x] Have circuit as selection when placed
- [x] Fix codebeat issues ("this is more of a suggestion" -Leon)
- [x] ~~Add button, probably under the same utilities menu as the generator, to organize the selected components using the same underlying organization algorithm. (although probably only if the organizer is improved) (saving this for a future issue I think)~~ current algorithm for it is borderline a hack and should be removed when a better one is implemented
- [x] Condense NAND/NOR gates
- [x] ~~Improve error messages to be more descriptive to the user (Likely ignoring this since it is too vague)~~ error messages could be better, but they also could be a lot worse

Low Priority
- [x] ~~Improve placement algorithm (may be a higher priority if I had any idea how to do this, I am going to try a variation that groups by "lowest" "depth" rather than "highest" "depth" or this pdf: https://www.graphviz.org/Documentation/TSE93.pdf)~~ future issue
- [x] ~~Select output component (low priority because led is the only output with a single in port)~~ no need
- [x] ~~Change input component for each input~~ maybe something for later
- [x] ~~Allow for user to enter custom operator precedence ordering (would require modifying the underlying algorithm, but may not actually be that hard to do)~~ maybe for later
- [x] Refactor ExpressionParser.test.ts to be less repetitive

No Priority
- [x] ~~Fix the issue present in #468 and other related issues, currently a few tests are commented because they otherwise fail due to what is probably the same issue. Fortunately, these don't seem to cause issues when placed on the real circuit. One example expression is "!!a"~~ already a different issue

Currently, I view High Priority issues as those I must finish before considering changing this from a draft pr to an actual pr. Medium I think I could realistically also finish, and low/no priority issues as those I probably will not fix, and could be made separate issues after this pull request is made. While I do have tests for the underlying algorithm, I have not yet extensively tested the UI side of things. It is possible more bugs will popup in the meantime from that. For the time being, I'm open to input on whatever else should be added to these checklists, or any bugs anyone else finds. The branch is expression-to-circuit-v2


Xystem4 (commented):
It looks like there's some minor inconsistencies across the different notation options. The "Programming 1" symbols ( | , & , ^ , ! ) are all considered valid operators regardless of the notation style selected (so for instance, a | b works the same if Programming 1 is selected or if Literal 1 is selected). All other operators ( && , || , * , + , _ , and the words) are only considered valid if their notation is selected (so if I select Programming 1 then a valid statement is: "XOR & not", with the two inputs named XOR and not.

Also, the XOR and xor operators are not considered valid operators in _any_ notation set. It does however give a different error message when the correct notation is selected, which leads me to believe that the parsing is reading the "or" first, and then considering the "x" to be a variable name, rather than part of the operator.

One more minor thing, regardless of which notation is selected, the title of the IC generated will be using Programming 1 notation. My expectation would have been for the title to basically just be the text that the user input (maybe with the spacing made more consistent, or some other minor tweaks).

TGCrystal:
@Xystem4 
Should all be fixed now. The previous system just did a simple find and replace for the different formats, which caused all of those issues. The new system is more robust and should fix all of that, but let me know if you find any other issues!

Xystem4 (dismissed):
All the issues I previously pointed out seem to be fixed! All operators seem to properly be reserved if and only if they are contained within the selected notation option.

One suggestion I might propose would be including the "nand", "nor", and "xnor" operations for the literal notation options (personally, the usefulness of those operations over having to specify a not gate with parentheses is probably the only reason I would use the literal notation).

I also think it might make sense to have the Input Component Type be chosen with checkboxes (similar to the Notation) rather than a dropdown, for consistency's sake (it also takes one less click and all options are viewable anytime). I'm sure you had reasons though for the current style, and this is all subjective.

Those suggestions aside, I can't find any bugs or other issues that should prevent this from being merged.

src/site/pages/digital/src/containers/ExprToCircuitPopup/index.scss
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    .cancel {
+        background-color: $danger-color;
+        margin-right: 10px;
+
+        &:enabled:hover {
+            background-color: darken($danger-color, 20%);
+        }
+    }
```
LeonMontealegre, line 52: All subclasses need to be like

`&__subclassname`

the `&` inherits the parent's name, i.e. the parent in this case is `.exprToCircuitPopup`, so the child would be come `.exprToCircuitPopup__generate` or `.exprToCircuitPopup__cancel`.

We do this because otherwise, it infects the global classname space, especially with common names like "cancel". 
If some other file had a classname called "cancel" for some cancel button, it would inherit the same styles as this one which could be very confusing.

Could be worth having a global `.cancel` class though, in something like
https://github.com/OpenCircuits/OpenCircuits/blob/master/src/site/pages/digital/src/containers/App/index.css
or maybe
https://github.com/OpenCircuits/OpenCircuits/blob/master/src/site/shared/components/Popup/index.scss

src/site/pages/digital/src/containers/ExprToCircuitPopup/index.scss
```diff
@@ -0,0 +1,115 @@
...
+            border-color: #5ccee7;
+
+            &:active {
+                background-color: #4ba9bd;
+                border-color: #4ba9bd;
+            }
+        }
+    }
+
+    .exprToCircuitPopupSettings {
```
LeonMontealegre, line 92: See above comment

src/site/pages/digital/src/containers/ExprToCircuitPopup/index.scss
```diff
@@ -0,0 +1,115 @@
...
+        .exprToCircuitPopupNotation {
+            flex-grow: 1;
+        }
+
+        .exprToCircuitPopupOptions {
+            flex-grow: 1;
+        }
+    }
+
+    .errorMessage {
```
LeonMontealegre, line 111: See above comment

src/site/pages/digital/src/containers/ExprToCircuitPopup/index.tsx
```diff
@@ -0,0 +1,217 @@
...
+                            <label>Input Component Type:  </label>
+                            <select id="input"
+                                    value={input}
+                                    onChange={e => setInput(e.target.value)}
+                                    onBlur={e => setInput(e.target.value)}>
+                                <option key="Constant Low" value="Constant Low">Constant Low</option>
+                                <option key="Constant High" value="Constant High">Constant High</option>
+                                <option key="Button" value="Button">Button</option>
+                                <option key="Switch" value="Switch">Switch</option>
+                                <option key="Clock" value="Clock">Clock</option>
```
LeonMontealegre, line 188: I don't think you need a `key` here since it isn't in a loop, however according to my comment above, you might want to turn this into a loop but either way the key will be dynamically set

src/site/pages/digital/src/containers/ExprToCircuitPopup/index.tsx
```diff
@@ -0,0 +1,217 @@
...
+                    }}>Generate</button>
+
+                    <button className="cancel" type="button" onClick={() => {
+                        setExpression({ expression: "" });
+                        setErrorMessage({ errorMessage: "" });
+                        CloseHeaderPopups();
+                    }}>Cancel</button>
+
+                </div>
+                </Popup>
```
LeonMontealegre, line 213: Something looks funky with the spacing here

src/site/shared/containers/Header/Right/UtilitiesDropdown.tsx
```diff
@@ -0,0 +1,55 @@
+import {connect} from "react-redux";
+
+import {SharedAppState} from "shared/state";
+import {HeaderMenus, HeaderPopups} from "shared/state/Header/state";
+import {OpenHeaderMenu, OpenHeaderPopup, CloseHeaderMenus} from "shared/state/Header/actions";
+
+import {Dropdown} from "./Dropdown";
+
```
LeonMontealegre, line 8: Minor: Two blank lines b/t imports and code

src/site/pages/digital/src/containers/DigitalHeader/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+import {Dropdown} from "shared/containers/Header/Right/Dropdown";
+
+import {CircuitInfoHelpers} from "shared/utils/CircuitInfoHelpers";
+
+type Utility = {
+    popupName: HeaderPopups;
+    img: string;
+    text: string;
+}
+
```
LeonMontealegre, line 12: This type is defined in `UtilitiesDropdown`, don't want to redefine it

src/site/pages/digital/src/containers/DigitalHeader/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+function getExtraUtilities(): Utility[] {
+    return [
+        { 
+            popupName: "expr_to_circuit",
+            img: "img/icons/bool_expr_input_icon.svg",
+            text: "Boolean Expression to Circuit"
+        }
+    ];
+}
```
LeonMontealegre, line 21: You can just define 
```ts
const ExtraUtilties = [
        { 
            popupName: "expr_to_circuit",
            img: "img/icons/bool_expr_input_icon.svg",
            text: "Boolean Expression to Circuit"
        }
    ];
```

and then the type is implicit

src/site/shared/containers/Header/Right/UtilitiesDropdown.tsx
```diff
@@ -0,0 +1,55 @@
...
+}
+
+type Props = StateProps & DispatchProps & ExtraUtilitiesProps & OwnProps;
+const _UtilitiesDropdown = ({ curMenu, openMenu, openPopup, closeMenus, extraUtilities }: Props) => (
+    <Dropdown open={(curMenu === "tools")}
+              onClick={() => openMenu("tools")}
+              onClose={() => closeMenus()}
+              btnInfo={{title: "Utilities", src: "img/icons/tools.svg"}}>
+        {extraUtilities.map(tool => (
+            <div key={tool.popupName} onClick={() => { closeMenus(); openPopup(tool.popupName); }}>
```
LeonMontealegre, line 35: Only need `key` for dynamic elements that are created in loops

src/site/shared/containers/Header/Right/UtilitiesDropdown.tsx
```diff
@@ -0,0 +1,55 @@
...
+                <img src={tool.img} height="100%" alt="" />
```
LeonMontealegre, line 36: Don't leave "alt"s blank. We're not great at being accessible but it's a small step lol

src/site/shared/containers/Header/Right/index.tsx
```diff
@@ -1,24 +1,35 @@
...
 import {TutorialDropdown} from "./TutorialDropdown";
 import {OpenFileButton} from "./OpenFileButton";
 import {DownloadMenuDropdown} from "./DownloadMenuDropdown";
+import {UtilitiesDropdown} from "./UtilitiesDropdown";
 import {SignInOutButtons} from "./SignInOutButtons";
 import {SettingsMenu} from "./SettingsMenu";
 
 import "./index.scss";
 
+type Utility = {
```
LeonMontealegre, line 13: Minor: two blank lines b/t imports and code

src/site/shared/containers/Header/Right/index.tsx
```diff
@@ -1,24 +1,35 @@
...
+    popupName: HeaderPopups;
+    img: string;
+    text: string;
+}
```
LeonMontealegre, line 17: Also, again, we don't want to redefine the `Utility` type.

src/site/shared/containers/Header/Right/index.tsx
```diff
@@ -1,24 +1,35 @@
...
 
 type Props = {
     helpers: CircuitInfoHelpers;
+    extraUtilities: Utility[];
```
LeonMontealegre, line 21: You can export the `Utility` type from `UtilitiesDropdown` and reference it from that

src/site/shared/containers/Header/index.tsx
```diff
@@ -1,16 +1,23 @@
...
 import {HeaderLeft} from "./Left";
 import {HeaderRight} from "./Right";
 
 import "./index.scss";
 
+type Utility = {
+    popupName: HeaderPopups;
+    img: string;
+    text: string;
+}
```
LeonMontealegre, line 13: See above

KevinMackenzie (changes requested):
I'm looking at the ExpressionParser.ts code and the work you've done is very much on the right tracks.  But, there appears to be a lot of duplicated code in `parseXor`,`parseAnd`, etc, that should be factored into a generic `parse[Binary]Op` function.  There are also a few things relating to organization/structure that makes this de-duplication easier.

In general, it is good to organize parsing and code generation in stages with explicitly-defined intermediate representations.  Not only does this makes testing and debugging easier, but it also allows you to support different features at each representation.

For instance, the lexer (`GenerateTokens`) could output a list of `Token`s, where `Token` is a union type of `or`, `not`, `open`, `close`, `Input`, etc.  This intermediate format would be fed into a stage that verifies semantic correctness (i.e. matched parentheses) and outputs a tree representation of the expression.  Since boolean expressions are context-free, this intermediate format (`Token` list) is probably OK to omit.

So as an alternative, you could parse the input string into a tree data structure directly using some recursive method (since it is context free) that respects operator precedence.  When you have this tree well-defined, your generation code does not need to be concerned with the validity of its input.

Regardless of which method you use, this expression tree would be fed into the circuit generator (currently `parseXX`), which could traverse this tree bottom-up, construct appropriate gates, place them into the circuit, etc.  Since operator sequence is implied by the tree, all you need to do is have some `connectGate(gate, child1[, child2])` function that creates the wires and connects the ports for whichever gate got constructed.

The expression tree is also a good place to modify the structure of the expression, i.e. make the simplification from `not (x and y)` to `x nand y`.  This is also the form you would use to "simplify" the expression.

And just because you parse expressions as context-free doesn't mean you can't add context later.  For instance, `a + b * c, a = d+e+f, b=g*h*i` can be parsed as context-free. Split at the comma and each definition is parsed into separate trees.  A final pass connects references into a DAG that can be fed into an unmodified circuit generator.  Cycles can be detected, and all undefined variables become inputs.

src/app/core/utils/math/Graph.ts
```diff
@@ -97,4 +97,76 @@ export class Graph<V, E> {
...
+                    }
+                }
+            }
+            currentLayer = nextLayer;
+            nextLayer = [];
+        }
+
+        return ret;
+    }
+
```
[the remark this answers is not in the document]
TGCrystal: Condensed down so that they both call a common privated function. Also, I think the bfs function currently in the Graph class is actually a depth first search, but since it only indicates to the caller whether a node is found or not I don't think it would effectively make a difference. Either way, I decided not to modify it since it also assumes the graph is undirected, which for my purpose it isn't (but it is when isConnected() is called).

src/site/pages/digital/src/containers/DigitalHeader/index.tsx
TGCrystal: Changed to the const declaration rather than the function returning, but I still had to specify the Utility[] since VS Code didn't seem to realize that it was a Utility type.

src/site/shared/containers/Header/Right/UtilitiesDropdown.tsx
TGCrystal: Removing the key attribute results in an error in the chrome developer console:

index.js:1 Warning: Each child in a list should have a unique "key" prop.

Check the render method of `_UtilitiesDropdown`. See https://reactjs.org/link/warning-keys for more information.
    at div
    at _UtilitiesDropdown (http://localhost:3000/static/js/main.chunk.js:14014:3)
    at ConnectFunction (http://localhost:3000/static/js/vendors~main.chunk.js:49834:75)
    at div
    at HeaderRight (http://localhost:3000/static/js/main.chunk.js:14193:3)
    at header
    at Header (http://localhost:3000/static/js/main.chunk.js:14357:3)
    at DigitalHeader (http://localhost:3000/static/js/main.chunk.js:17325:3)
    at main
    at div
    at Provider (http://localhost:3000/static/js/vendors~main.chunk.js:49547:20)

src/site/pages/digital/src/containers/ExprToCircuitPopup/index.scss
TGCrystal: Moved .cancel to https://github.com/OpenCircuits/OpenCircuits/blob/master/src/site/shared/components/Popup/index.scss

src/app/core/utils/ComponentOrganizers.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    // Convert the map to a 2d array where index of first array is the depth
+    for (const [index, depth] of depthMap.entries()) {
+        while (depthArray.length <= depth)
+            depthArray.push([]);
+        depthArray[depth].push(components[index]);
+    }
+
+    const xOffset: number = 250;
+    const yOffset: number = 150;
```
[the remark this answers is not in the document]
TGCrystal: They are arbitrary values used to set the spacing between components. I pretty much just assigned them some random values until I found what I thought looked good enough.
The start vector that is passed in is what is used to define the top left corner where the circuit placement starts. This value in ExprToCircuitPopup is defined as `info.camera.getPos().sub(info.camera.getCenter().scale(info.camera.getZoom()/1.5));` which is intended to be near the top left corner of the camera's view, since the circuit placement expands down and to the right.

As for the second part, I don't seem to be having that issue. If I move to a different part of the scene and even if I change the zoom level, the circuit still starts to be placed in the top left corner of the camera's view.

src/app/core/utils/ComponentOrganizers.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    for (let i = 0; i < depthArray.length; i++) {
+        for (let j = 0; j < depthArray[i].length; j++) {
+            const component = depthArray[i][j];
+            const newLocation = new Vector(start.x + xOffset * i, start.y + yOffset * j);
+            component.setPos(newLocation);
+        }
+    }
+}
```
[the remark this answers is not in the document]
TGCrystal: I agree this whole thing is incredibly convoluted, should I try work on refactoring it or is there some other approach that I should check out instead?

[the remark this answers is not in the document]
LeonMontealegre: Okay, so I took a closer look and I believe I know what you're trying to do.
Basically, you created this Map of Nodes to their Depth, and then in this method, you are basically finding the inverse, i.e., the map of each Depth to every Node at that depth, and then using that information to place it.

I believe the best course of action here is to actually take in a `Map<number, number[]>` and changed `Graph.getNodeDepths` to find the Mapping of depths to nodes instead of nodes to depths.

This would drastically simplify the OrganizeCore method to something like:
```ts
function OrganizeCore(groups: IOObjectSet, start: Vector, depthMap: Map<number, number[]>): void {
    // Loop through each depth-level
    Array.from(depthMap.entries()).forEach(([depth, indices]) =>
        // Then loop through each component at each depth and set position
        //  with x-position related to `depth` and y-position related to `i`
        indices.map(c => groups.getComponents()[c])
               .forEach((c, i) => c.setPos(start.add(ORGANIZE_SEP_X*depth - depthMap.size/2, 
                                                     ORGANIZE_SEP_Y*i - indices.length/2)))
    );
}
```

Also, for the constants for `offset` you should instead make them constants in `core/utils/Constants` and call them something like I named them above (`ORGANIZE_SEP_X`, `ORGANIZE_SEP_Y`)


Also for below for `OrganizeMax/MinDepth`
you can 1-line them and have it just be `OrganizeCore(groups, start, CreateGraph(groups).getMaxNodeDepths())`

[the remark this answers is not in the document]
TGCrystal: Made these changes, but rather than have depthMap be a Map, I just used a 2d array where each "inner array" contains the components for a given depth, and the index of that array is the the depth that those components are at.

TGCrystal:
> Also, small "bug" - you can press generate even when there isn't anything written in the text box. The "generate" button should probably be disabled until the user has typed something

Now the generate button only shows up when something has been input. I could change it so it just grays out the button and nothing happens when it is clicked when the input is empty.

KevinMackenzie (changes requested):
These changes are definitely moving the in right direction, but your method of verifying the input correctness before generating the tree has a couple of problems.  First, its hard to follow how the verification you do relates to the assumptions you make in `generateTreeCore`.  Second, its hard to completely verify syntactic correctness without fully parsing the expression at the same time.

You want your data structures to maintain, by construction, the truths you verify when parsing.  A general-purpose graph isn't very good at that.  If you parse into something that represents the structure of the grammar like this (maybe not exactly this):
```ts
type InputTreeBinOpType = "|" | "^" | "&"
type InputTreeUnOpType = "!"

interface InputTreeIdent {
    kind: "leaf"
    ident: string
}
interface InputTreeUnOpNode {
    kind: "unop"
    type: InputTreeUnOpType
    child: InputTree
}
interface InputTreeBinOpNode {
    kind: "binop"
    type: InputTreeBinOpType
    lChild: InputTree
    rChild: InputTree
}
type InputTree =
    | InputTreeIdent
    | InputTreeUnOpNode
    | InputTreeBinOpNode
```
you know that whatever the parser returns is a valid expression and its trivial to convert to a general-purpose graph (albeit more verbose).  The actual parsing logic may be largely the same.

Also, parser code can be confusing, so please comment the logic in `generateTreeCore` : )

src/site/pages/digital/src/containers/ExprToCircuitPopup/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+type OwnProps = {
+    info: DigitalCircuitInfo;
+}
+type StateProps = {
+    curPopup: HeaderPopups;
+}
+type DispatchProps = {
+    CloseHeaderPopups: typeof CloseHeaderPopups;
+}
```
LeonMontealegre, line 46: We recently refactored how to pass props/actions from Redux, take a look at the other containers and it should be fairly straightforward and make the code much nicer!

src/site/shared/containers/Header/Right/UtilitiesDropdown.tsx
```diff
@@ -0,0 +1,56 @@
...
+    curMenu: HeaderMenus;
+}
+type DispatchProps = {
+    openMenu: (menu: HeaderMenus) => void;
+    openPopup: (popup: HeaderPopups) => void;
+    closeMenus: () => void;
+}
+type ExtraUtilitiesProps = {
+    extraUtilities: Utility[];
+}
```
LeonMontealegre, line 27: See above about Redux refactor

LeonMontealegre (changes requested):

src/site/pages/digital/src/containers/ExprToCircuitPopup/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+        const inputTypes = Array.from(Inputs.keys()).map((input) =>
+            <option key={input} value={input}>{input}</option>
+        );
+    
+        const formats = Array.from(FormatMap.entries()).map((formatEntry) =>
+            <div key={formatEntry[0]}>
+                <input type="radio" id={formatEntry[0]} name="format" checked={format === formatEntry[0]} onChange={() => setFormat(formatEntry[0])} value={formatEntry[0]} />
+                <label htmlFor={formatEntry[0]}>{formatEntry[1].get("label")}</label><br/>
+            </div>
+        );
```
LeonMontealegre, line 115: These should just be directly in the returned React code

If they're needed in multiple places, make a separate component for them like the custom ops below

src/site/pages/digital/src/containers/ExprToCircuitPopup/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+            ["(", "("],
+            [")", ")"],
+            ["Separator", "separator"],
+        ];
+        const customOpsInput = customOps.map((op) => 
+            <div className="exprtocircuit__popup__customOps" key={op[0]}>
+                <input title={"Enter symbol for " + op[0]} type="text" value={userOps[op[1]]} onChange={e => setUserOps({...userOps, [op[1]]: e.target.value})}/>
+                <label htmlFor={userOps[op[1]]}>Custom {op[0]}: "{userOps[op[1]]}"</label>
+            </div>
+        );
```
LeonMontealegre, line 131: This can be its own React component called like `CustomOps` (same folder, new file `CustomsOps.tsx`)

The state `userOps` can stay here, and the `CustomOps` component can just take whatever it needs as props

src/site/pages/digital/src/containers/ExprToCircuitPopup/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+        return (
+            <Popup title="Digital Expression To Circuit Generator"
+                   isOpen={(curPopup === "expr_to_circuit")}
+                   close={() => dispatch(CloseHeaderPopups())}>
+                <div className="exprtocircuit__popup">
+                    { errorMessage && <p className="exprtocircuit__popup__errorMessage">{"ERROR: " + errorMessage}</p> }
+                    <input title="Enter Circuit Expression" type="text"
+                               value={expression}
+                               placeholder="!a | (B^third)"
+                               onChange={e => setExpression(e.target.value)} />
```
LeonMontealegre, line 142: nit: align the properties

src/site/pages/digital/src/containers/ExprToCircuitPopup/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+                            <h3>Options</h3>
+                            <input onChange={() => setIsIC(!isIC)} checked={isIC} type="checkbox" id="isIC" name="isIC" />
+                            <label htmlFor="isIC">Generate into IC</label>
+                            <br/>
+                            <br/>
+                            <label>Input Component Type:  </label>
+                            <select id="input"
+                                value={input}
+                                onChange={e => setInput(e.target.value)}
+                                onBlur={e => setInput(e.target.value)}>
```
LeonMontealegre, line 171: nit: align properties

src/site/pages/digital/src/containers/ExprToCircuitPopup/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+                            {inputTypes}
```
LeonMontealegre, line 172: nit: indent 1 more to the right

src/site/pages/digital/src/containers/ExprToCircuitPopup/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+                                format === "custom" &&
+                                <div>
+                                    {customOpsInput}
+                                </div>
+                            }
+                        </div>
+
+                        <div>
+                            <h3>Options</h3>
+                            <input onChange={() => setIsIC(!isIC)} checked={isIC} type="checkbox" id="isIC" name="isIC" />
```
LeonMontealegre, line 163: nit: id should be first property, followed by `type` then `name`, `onChange` should be last

src/site/pages/digital/src/containers/ExprToCircuitPopup/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+                    </div>
+
+                    <button className="exprtocircuit__popup__generate" type="button" disabled={expression===""} onClick={() => {
+                        try {
+                            generate(mainInfo, expression, isIC, input, format, new Map(Object.entries(userOps) as [FormatLabels, string][]));
+                            setExpression("");
+                            setErrorMessage("");
+                            dispatch(CloseHeaderPopups());
+                        }
+                        catch (err) {
```
LeonMontealegre, line 185: nit: catch can be on same line as `}`

src/site/pages/digital/src/containers/ExprToCircuitPopup/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+                            setErrorMessage(err.message);
+                        }
+                    }}>Generate</button>
+
+                    <button className="cancel" type="button" onClick={() => {
+                        setExpression("");
+                        setErrorMessage("");
+                        dispatch(CloseHeaderPopups());
```
LeonMontealegre, line 193: You can declare like a `reset` function above (in the component still though) since this is being used twice

src/site/pages/digital/src/containers/ExprToCircuitPopup/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+import {LED}                    from "digital/models/ioobjects/outputs/LED";
+import {ConstantLow}            from "digital/models/ioobjects/inputs/ConstantLow";
+import {ConstantHigh}           from "digital/models/ioobjects/inputs/ConstantHigh";
+import {Button}                 from "digital/models/ioobjects/inputs/Button";
+import {Switch}                 from "digital/models/ioobjects/inputs/Switch";
+import {Clock}                  from "digital/models/ioobjects/inputs/Clock";
+import {ExpressionToCircuit}    from "digital/utils/ExpressionParser";
+import {GenerateTokens} from "digital/utils/ExpressionParser/GenerateTokens";
+import {FormatLabels} from "digital/utils/ExpressionParser/Constants/DataStructures";
+import {FormatMap} from "digital/utils/ExpressionParser/Constants/Objects";
```
LeonMontealegre, line 32: nit: utils should be above models and with the other `digital/utils` import and they should probably align the `from`s with eachother

src/site/pages/digital/src/containers/ExprToCircuitPopup/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+import {DigitalCircuitInfo}     from "digital/utils/DigitalCircuitInfo";
+import {DigitalComponent}       from "digital/models/DigitalComponent";
+import {ICData}                 from "digital/models/ioobjects/other/ICData";
+import {IC}                     from "digital/models/ioobjects/other/IC";
+import {LED}                    from "digital/models/ioobjects/outputs/LED";
+import {ConstantLow}            from "digital/models/ioobjects/inputs/ConstantLow";
+import {ConstantHigh}           from "digital/models/ioobjects/inputs/ConstantHigh";
+import {Button}                 from "digital/models/ioobjects/inputs/Button";
+import {Switch}                 from "digital/models/ioobjects/inputs/Switch";
+import {Clock}                  from "digital/models/ioobjects/inputs/Clock";
```
LeonMontealegre, line 28: nit: Put a space b/t this import and from `utils`, the models can be vertically aligned just with themself

src/site/pages/digital/src/containers/ExprToCircuitPopup/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+import {useSharedDispatch, useSharedSelector} from "shared/utils/hooks/useShared";
+
+import {OrganizeMinDepth} from "core/utils/ComponentOrganizers";
+
+import {GroupAction}             from "core/actions/GroupAction";
+import {CreateDeselectAllAction,
+        CreateGroupSelectAction,
+        SelectAction,}            from "core/actions/selection/SelectAction";
+import {PlaceAction}             from "core/actions/addition/PlaceAction";
+import {CreateICDataAction}      from "digital/actions/CreateICDataAction";
```
LeonMontealegre, line 17: nit: extra line above this

src/site/shared/containers/Header/Right/index.tsx
```diff
@@ -24,6 +27,9 @@ export const HeaderRight = ({ helpers }: Props) => {
             </button>
             <div className={`header__right__btns ${isHidden ? "header__right__collapsed" : ""}`}>
                 <SignInOutButtons />
+                {extraUtilities.length > 0 && // Render only if there are tools
```
LeonMontealegre, line 30: nit: comment wrong

src/site/pages/digital/src/containers/DigitalHeader/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+    }
+];
+
+type Props = {
+    img: string;
+    helpers: CircuitInfoHelpers;
+}
+
+export const DigitalHeader = ({ img, helpers }: Props) => (
+    <Header img={img} helpers={helpers} extraUtilities={ExtraUtilties} />
```
LeonMontealegre, line 21: You can just put the `ExtraUtilities` inline, this function is same enough to warrant that

src/app/digital/utils/ExpressionParser/TreeToCircuit.ts
```diff
@@ -0,0 +1,71 @@
...
+ * @param node the root node of the InputTree to convert
+ * @param inputs the input components used by this expression
+ * @param circuit used to store the circuit while recursing, on the intial call it should contain the DigitalComponents found in inputs
+ * @returns the current part of the tree that has been converted to a circuit, the most recently used component
+ *              should always be last in the array
+ * @throws {Error} when one of the leaf nodes of the InputTree references an input that is not inputs
+ * @see TreeToCircuit
+ */
+ function treeToCircuitCore(node: InputTree, inputs: Map<string, DigitalComponent>, circuit: IOObject[]): IOObject[] {
+    if (node.kind === "leaf") { //Rearranges array so thge relevant input is at the end
```
LeonMontealegre, line 23: nit: space after `//`

src/app/digital/utils/ExpressionParser/GenerateTokens.ts
```diff
@@ -0,0 +1,93 @@
...
+    if (ops.get("separator") === "")
+        throw new Error("Length zero separator in supplied operation symbols");
+
+    const tokenList = new Array<Token>();
+    let extraSkip = 0;
+    let token: Token;
+
+    let index = 0;
+
+    while(index < expression.length) {
```
LeonMontealegre, line 74: nit: space after while

src/app/digital/utils/ExpressionParser/GenerateTokens.ts
```diff
@@ -0,0 +1,93 @@
...
+        if (token === null)
+            index += ops.get("separator").length;
+        else if (token.type === "input") {
+            tokenList.push(token);
+            index += token.name.length;
+        }
+        else {
+            tokenList.push(token);
+            index += ops.get(token.type).length;
+        }
```
LeonMontealegre, line 87: nit: don't mix and match one-liner no-brace-usage with else-ifs that have braces, just have braces for each statement

nit: end of if `}` can have the elseif/else on the same line

src/app/digital/utils/ExpressionParser/GenerateTokens.ts
```diff
@@ -0,0 +1,93 @@
...
+        if (ops.get(op) === "")
+            throw new Error("Length zero " + op + " in supplied operation symbols");
+    }
+    if (!ops.has("separator"))
+        throw new Error("No separator in supplied operation symbols");
+    if (ops.get("separator") === "")
+        throw new Error("Length zero separator in supplied operation symbols");
+
+    const tokenList = new Array<Token>();
+    let extraSkip = 0;
```
LeonMontealegre, line 69: I don't think `extraSkip` gets set to anything but 0

src/app/digital/utils/ExpressionParser/GenerateTokens.ts
```diff
@@ -0,0 +1,93 @@
...
+ * Gets a token from a given expression starting at a certain index
+ * 
+ * @param expression the expression to extract the token from
+ * @param index the index where the token starts
+ * @param ops the representation format for the operations used in this expression
+ * @returns the token extracted from the expression or null if the index points to the starting location of
+ *              a separator (like " ")
+ */
+function getToken(expression: string, index: number, ops: Map<FormatLabels, string>): Token | null {
+    for(const op of OpsArray) {
```
LeonMontealegre, line 38: nit: space after for

src/app/digital/utils/ExpressionParser/GenerateTokens.ts
```diff
@@ -0,0 +1,93 @@
...
+ * 
+ * @param expression the expression to extract the input name from
+ * @param index the index at which the input starts
+ * @param ops the representation format for the operations used in this expression
+ * @returns an InputToken with the input name in it
+ */
+ function getInput(expression: string, index: number, ops: Map<FormatLabels, string>): InputToken {
+    let endIndex = index + 1;
+    while(endIndex < expression.length) {
+        for(const op of OpsArray) {
```
LeonMontealegre, line 17: nit x2: space after while/for

src/app/digital/utils/ExpressionParser/GenerateInputTree.ts
```diff
@@ -0,0 +1,144 @@
...
+            const nextToken = tokens[nextIndex];
+            if (nextToken.type === "input") {
+                next = nextToken.name;
+                break;
+            }
+        }
+        if (prev !== null && next !== null)
+            throw new Error("No valid operator between \"" + prev + "\" and \"" + next + "\"");
+        
+        console.log(ret.tree);
```
LeonMontealegre, line 139: console log

src/app/digital/utils/ExpressionParser/GenerateInputTree.ts
```diff
@@ -0,0 +1,144 @@
...
+        let prev: string = null;
+        for(let prevIndex = index-1; prevIndex >= 0; prevIndex--) {
+            const prevToken = tokens[prevIndex];
+            if (prevToken.type === "input") {
+                prev = prevToken.name;
+                break;
+            }
+        }
+        let next: string = null;
+        for(let nextIndex = index; nextIndex < tokens.length; nextIndex++) {
```
LeonMontealegre, line 129: nit: space after for

src/app/digital/utils/ExpressionParser/GenerateInputTree.ts
```diff
@@ -0,0 +1,144 @@
...
+        return null;
+    const ret = generateInputTreeCore(tokens, ops);
+
+    const index = ret.index;
+    if (index < tokens.length) {
+        if (tokens[index].type === ")")
+            throw new Error("Encountered Unmatched \"" + ops.get(")") + "\"");
+
+        let prev: string = null;
+        for(let prevIndex = index-1; prevIndex >= 0; prevIndex--) {
```
LeonMontealegre, line 121: nit: space after for

src/app/digital/utils/ExpressionParser/GenerateTokens.ts
```diff
@@ -0,0 +1,93 @@
+import {SubEquals} from "core/utils/StringUtils";
```
LeonMontealegre, line 1: In general see if any of these `for` loops can be reduced to like `find`, `findIndex`, `map`, etc.

src/app/digital/utils/ExpressionParser/GenerateInputTree.ts
```diff
@@ -0,0 +1,144 @@
...
+ * @throws {Error} parenthesis do not include anything (such as "()")
+ * @throws {Error} an opening parenthesis is missing a corresponding closing parenthesis (such as "(")
+ * @throws {Error} a closing parenthesis is missing a corresponding opening parenthesis (such as ")")
+ * @throws {Error} |, &, or ^ are missing an operand on their left (such as "a|")
+ * @throws {Error} |, &, ^, or ! are missing an operand on their right (such as "!a")
+ * @throws {Error} there is no operator between two inputs (such as "a b")
+ * @throws {Error} generateInputTreeCore returns back up to this function before the end of tokens is reached
+ *                  for any other reason
+ */
+export function GenerateInputTree(tokens: Array<Token>, ops: Map<FormatLabels, string> = FormatMap.get("|")): InputTree | null {
```
LeonMontealegre, line 110: In general see if any of these for loops can be reduced to like `map`, `find`, `some`, `every`

src/app/digital/utils/ExpressionParser/GenerateInputTree.ts
```diff
@@ -0,0 +1,144 @@
...
+        rightRet = generateInputTreeCore(tokens, ops, currentOp, index);
+    else if (currentOp === "!" && rightToken.type === "input") // This case would apply when an input follows a "!"
+        rightRet = {index: index+1, tree: {kind: "leaf", ident: rightToken.name}};
+    else if (currentOp === "(") {
+        if (index >= tokens.length)
+            throw new Error("Encountered Unmatched \"" + ops.get("(") + "\"");
+        rightRet = generateInputTreeCore(tokens, ops, nextOp, index);
+    }
+    else
+        rightRet = generateInputTreeCore(tokens, ops, currentOp, index);
```
LeonMontealegre, line 74: nit: Don't mix braces and non-braces, also can keep elseif/elses on same line as ending `}`

src/app/digital/utils/ExpressionParser/GenerateInputTree.ts
```diff
@@ -0,0 +1,144 @@
...
+ * @throws {Error} |, &, ^, or ! are missing an operand on their right (such as "!a")
+ * @see GenerateInputTree
+ */
+ function generateInputTreeCore(tokens: Array<Token>, ops: Map<FormatLabels, string>, currentOp: TokenType = "|", index: number = 0): NewTreeRetValue {
+    const nextOp = DefaultPrecedences.get(currentOp);
+    if (tokens[index].type === ")") {
+        if (index > 0) {
+            if (tokens[index-1].type === "(")
+                throw new Error("Empty Parenthesis");
+            else if (IsOperator(tokens[index-1]))
```
LeonMontealegre, line 27: nit: this can just be another `if` since the previous `if` throws and therefore stops execution

src/app/digital/utils/ExpressionParser/GenerateInputTree.ts
```diff
@@ -0,0 +1,144 @@
...
+    // When this function has recursed through to "!" and the token still isn't that, then
+    //  the only possibilites left are an input or open parenthesis token
+    if (currentOp === "!" && tokens[index].type !== "!") {
+        const token = tokens[index];
+        if (token.type === "(")
+            return generateInputTreeCore(tokens, ops, nextOp, index);
+        else if (token.type === "input")
+            return {index: index+1, tree: {kind: "leaf", ident: token.name}};
+        else
+            throw new Error("Missing Left Operand: \"" + ops.get(token.type) + "\"");
```
LeonMontealegre, line 42: nit: don't need elseif/elses since the bodies of the statements return/throw

LeonMontealegre (changes requested):

src/site/pages/digital/src/containers/ExprToCircuitPopup/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+import {CloseHeaderPopups} from "shared/state/Header";
+import {useSharedDispatch,
+        useSharedSelector} from "shared/utils/hooks/useShared";
+
+import {DigitalCircuitInfo}  from "digital/utils/DigitalCircuitInfo";
+
+import {Generate} from "./generate";
+
+import "./index.scss";
+import {CustomOps} from "./CustomOps";
```
LeonMontealegre, line 17: nit: import order

src/site/pages/digital/src/containers/ExprToCircuitPopup/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+                        <label htmlFor="isIC">Generate into IC</label>
+                        <br/>
+                        <br/>
+                        <label>Input Component Type:  </label>
+                        <select id="input"
+                                value={input}
+                                onChange={e => setInput(e.target.value as InputTypes)}
+                                onBlur={e => setInput(e.target.value as InputTypes)}>
+
+                            {(["Button", "Clock", "Switch"] as Array<InputTypes>).map(input =>
```
LeonMontealegre, line 90: nit: use `InputTypes[]`

LeonMontealegre (changes requested):

LeonMontealegre (approved):

PR #650, opened 2021-04-25
Create n components

### LINKED ISSUE #579: Quick Create N-components
It can be annoying to create, for instance, a bunch of switches and manually drag and drop a bunch of them in a neat way.

It would be nice to be able to click, for instance, a Switch, N times and then have N switches to drop somewhere onto the scene. They can all be in a column together and then placed with a single other click.

cchen1000:
Clicking on a component N times will drop the components, in a column, on the canvas by clicking on it. Nothing will drop if something else, other than the canvas is clicked, after clicking the components.

Demo:
demo

Fixes #579


Xystem4 (commented):
I'm not entirely sure if these were intended, but I'll point out that the spacing between the pasted in components scales with the user's zoom (on the default zoom LEDs, flip flops, and MUX/DeMUX overlap with one another), and that components will not be placed on the canvas if they would be placed somewhere outside the user's current view. Also, undoing will undo the placing one component at a time (that seems intentional though) (same with redo).

The only issue I've found that's definitely unintentional is an error that is thrown when trying to use this on an IC:

image

LeonMontealegre (approved):

PR #671, opened 2021-07-09
Issue 655: "Components Going Invisible when moving it towards the Edge of the site" addressed

Demopans:
placeholder pr until actual fix is found and implemented


LeonMontealegre (approved):
I figured it out! Turns out I had missed a single update for the camera's `dirty` optimization!!

PR #673, opened 2021-07-14
Dragging component moves component to front layer

### LINKED ISSUE #633: Dragging components should bring them to the front layer
**Describe the bug**
When dragging and moving a component, it remains at the exact layer (relative to other components) that it was before being dragged.

**To Reproduce**
1. Place multiple components onto the canvas
2. Position the components such that they are overlapping in the same region
3. Click and drag one of the components not on top

**Expected behavior**
We would expect a component to be brought to the front layer upon being dragged.

**Screenshots**
Three overlapping components
first
After dragging the constant high component
second
Expected result
third

 - OS: Windows 10
 - Browser: chrome

**Additional context**
Clicking and selecting components properly brings it to the front, it is only when dragging but _not_ selecting them that this behavior occurs

DefinitiveAbove:
### What is the current behavior?
Dragging a component does not change the layer that it is on.

### What is the new behavior?
Fixes #633 
Dragging a component brings it up to the front layer.


LeonMontealegre (approved):

LeonMontealegre (changes requested):
Taking a look at a different issue I realized - this currently creates 2 separate actions:
1. for moving to the front
2. for dragging
You should have it combine into 1 action at the end once the tool deactivates.

To do this you can run the ShiftAction w/o adding it to the history in `onActivate` and then create the action again and add it to the history w/ the TranslateAction in `onDeactivate`

LeonMontealegre (approved):

PR #676, opened 2021-07-16
Empty Actions Fix

### LINKED ISSUE #622: Clicking on empty space counts as an action (affecting undo/redo)
**Describe the bug**
When performing some amount of actions, then clicking anywhere on the screen a few times, then performing more actions, the history of actions includes each of those empty clicks as an action. Meaning that to undo to the point _before_ those empty clicks, you will need to undo several times in a row (with nothing happening)

**To Reproduce**
1. Add any component to the canvas
2. click on any blank area of the canvas several times
3. add any other new component
4. perform the undo action (ctrl+z or clicking the undo button, or cmd+z, etc.)
5. It will take many undo actions after removing the second component, before the first will be removed.

**Expected behavior**
We would expect empty clicks to not be logged as actions at all, aside from in the sense of when they cause a deselection (in which case it should only be logged as a single action).

 - OS: Windows 10
 - Browser: chrome

SamooMei:
Empty clicks were counted as a standalone action thus causing Undo/Redo to not work properly. Now empty clicks won't count as an action unless deselecting a component.

Fixes #622

### DIFF
--- a/src/app/core/tools/handlers/SelectionHandler.ts
+++ b/src/app/core/tools/handlers/SelectionHandler.ts
@@ -39,6 +39,7 @@ export const SelectionHandler: EventHandler = ({
             const deselect = (input.isShiftKeyDown() && selections.has(obj));
             action.add(new SelectAction(selections, obj, deselect).execute());
             action.add(new ShiftAction(designer, obj).execute());
+            history.add(action);
         }
     }
 });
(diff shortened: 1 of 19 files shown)


src/app/core/tools/handlers/SelectionHandler.ts
```diff
@@ -39,8 +39,7 @@ export const SelectionHandler: EventHandler = ({
             const deselect = (input.isShiftKeyDown() && selections.has(obj));
             action.add(new SelectAction(selections, obj, deselect).execute());
             action.add(new ShiftAction(designer, obj).execute());
+            history.add(action);
```
LeonMontealegre, line 42: i have a sneaking suspicion this is going to break deselect-ing, we do a `action.add` above that

please confirm

SamooMei: should be fixed

LeonMontealegre (approved):

LeonMontealegre (approved):

PR #686, opened 2021-08-02
ConnectionAction creates wire just-in-time to fix Snip action

### LINKED ISSUE #685: "Snip" tool is broken / causes errors
**Describe the bug**
Using the "snip" tool induces "cannot read _ of undefined" errors in the console and the site becomes broken / unresponsive.

**To Reproduce**
1. Add a source and sink component
2. Connect with a wire
3. Click wire and drag to create "node"
4. Select "node" and press "x" to "snip"
5. see console...

**Desktop**
 - OS: Linux
 - Browser: chrome
 - Version: 92

**Additional context**
It may be because the `CreateSnipWireAction` in `SplitWireAction.ts` creates a `ConnectionAction` that connects two ports that already are connected.  Since this wire is created before the `DisconnectAction`s are invoked, the `DigitalCircuitDesigner` refuses to create a wire and returns undefined.  Attempting to connect this undefined wire when invoking the `ConnectionAction` causes problems.

KevinMackenzie:
closes #685 

If the `ConnectionAction` is created before it is executed the `DigitalCircuitDesigner` may fail to create the wire, as explained in the issue.  Now, the wire is initialized once, just-in-time, so the precondition of adding the wire isn't enforced prematurely.


LeonMontealegre (changes requested):
This appears to still fail when snipping multiple ports at once

After a quick look it looks like the solution might be adding `.execute()` to each of the actions as they're added in the `CreateSnipWireAction` function (`src/app/core/actions/addition/SplitWireAction.ts`) as `CreateSplitWireAction` does

This thing you're doing rn is something that was abundant and I avoided like the devil in the most recent refactor so I am unfortunately not going to accept it

The reasons for the instant `.execute()` is a push towards eventually having all Actions execute immediately on creation w/in the constructor which I hope to do in the future

LeonMontealegre (approved):

PR #693, opened 2021-08-12
Sign in/out on SideNav

### LINKED ISSUE #688: Sign in/out button should also be in header of SideNav
Something like this

Screen Shot 2021-08-02 at 1 04 20 PM

SamooMei:
Description: SideNav now uses the Sign in/out component and the name now truncates itself when its too long

| Before     | After |
| ---      | ---       |
| Before |  After |

Fixes #688


KevinMackenzie:
The button works as described.  The sign-in/sign-out button still exists in the right side of the title bar.  Do we want the button in both places?

LeonMontealegre:
> The sign-in/sign-out button still exists in the right side of the title bar. Do we want the button in both places?

Yes, the more options the better!

LeonMontealegre (approved):

Issue #657, opened 2021-05-03
Coordinate default zoom with zoom-to-fit with no components


Xystem4:
Currently, if a user uses the zoom-to-fit functionality with nothing on the canvas, it will put the screen at a preset zoom and location. However, this level of zoom is different from the default zoom the screen is at when the page is first opened, which seems counterintuitive. I would expect both refreshing the page and pressing zoom-to-fit to place the screen at the exact same levels.

The default zoom upon opening the page:
![Screenshot (176)](https://user-images.githubusercontent.com/65842664/116924594-14a4a000-ac26-11eb-901c-f8e600d44713.png)
The zoom after using zoom-to-fit:
![Screenshot (177)](https://user-images.githubusercontent.com/65842664/116924604-1b331780-ac26-11eb-970f-0f0a20066135.png)

The second image appears to more precisely fit the size of the gridlines to the screen, so if I were to blindly suggest one of the above images to make the default, it would be the latter.

PR #720, opened 2021-09-29
fixed issue #657: Coordinate default zoom with zoom-to-fit with no co…

Kevin-771:
…mpnents


Kevin-771:
whoops didn't mean to push the boundingbox.ts here

PR #763, opened 2021-10-10
Esc and Enter for ICs

### LINKED ISSUE #713: Allow keyboard use for confirming/canceling IC creation
When in the IC creation menu, pressing enter should trigger the "Confirm" option, and pressing Escape should trigger the "Cancel" option (unless the IC Name text field is currently in focus, in which case pressing escape should un-focus from the text field, as per #632. Pressing enter should confirm regardless of text field focus).

This feels like a natural interaction, as escape already works similarly in other areas (color selection for LEDs and labels), and pressing enter to confirm just after typing in a name also seems intuitive.

thedomino1313:
Added functionality to confirm an IC creation with the enter key, and to cancel creation with the esc key, as well as closing the IC Viewer with esc. Added a keydown hook that will listen for a key press and execute a function when triggered.

Fixes  #713

### DIFF
--- a/src/site/pages/digital/src/containers/ICDesigner/index.tsx
+++ b/src/site/pages/digital/src/containers/ICDesigner/index.tsx
@@ -190,7 +190,7 @@ export const ICDesigner = (() => {
                         placeholder="IC Name"
                         onChange={(ev) => setName({name: ev.target.value})}
                         onKeyUp={(ev) => {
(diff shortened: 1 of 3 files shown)


OwenMcGee:
If possible, pressing enter should confirm IC creation when the "IC Name" textfield is in focus (currently enter only has an effect when the text field is _not_ in focus)

PR #768, opened 2021-10-13
Implemented loading bar/messages

### LINKED ISSUE #750: Status Bar/Messages on Loading Screen
Adding a loading progress bar to the loading screen would help to indicate to users (particularly those with slow internet connections or computers) that the site is indeed being loaded and not just looping forever.
Adding loading status messages could also be used to help debug issues if the site is actually stuck loading.

TGCrystal:
Closes #750 

This ended up being far more straightforward than I expected (I think). I'm open to change on any of the css and actual message contents. More progress points could probably be added as well. I'm not sure what the best way to actually step through the loading process to make sure the error message is properly changing at all points is, so that may need some testing.


TGCrystal:
I like the suggestions, I made some changes and I do have a few further clarifying questions. 

1. The image code looks like: 
```
Load: async function(): Promise<void> {
    const promises =
        IMAGE_FILE_NAMES.map(async (name) =>
            await loadImage(name)
        );

    await Promise.all(promises);
}
```
I'm not super familiar with async stuff, but if I understand it currectly, this is synchronously starting the loading process for images, but loading each image is asynchronous. If that understanding is correct, how can I update when an image finishes loading, not just starting to load?

2. What scripts other than the Google Auth script are there to load? I would ideally like to have the text say what script is currently being loaded. I first opened this issue because I was having some weird interference with an adblocker on my phone that was preventing the site from loading, so having that level of debug messaging could be helpful in conjunction with a link to the GitHub. For now, I added the part that mentions it may take awhile on slow networks to the initial message. 

3. This is kinda just a restatement of 2, but what exactly happens between the initial loading text appearing and Init being called?

LeonMontealegre:
1. So basically each image is mapped to a "promise" that waits for that image to load. This is done via the map function to a bunch of async functions, right now it's all compressed and inlined but the map takes in a function so it can first `await loadImage(name)` and then do something after that which would be when that individual image loads, a counter could be kept outside that keeps track of the number of loaded images and then call the callback with that (don't worry about usual async problems with multiple threads changing a single variable because JS is single threaded and "async" is kind of a lie kinda it's complicated but also makes it simple)

2. When i say script i mean the entire codebase, all of the code starting from index.tsx is bundled into one mega-script that's 3.6MB in dev mode so it takes a long time to load in throttled mode. That's why the loading screen has to be in pure HTML/CSS because if we used React we'd have to wait for the entire bundle to load because React is just JavaScript at the end of the day

3. ^

TGCrystal:
I now have the loading bar increment when images are loading (and there is a counter displayed for how many images have loaded). The message that loading initial scripts may take awhile now shows up after a few seconds rather than right away. I put try catch blocks around a bunch of parts of Init. If an error is caught, a summary is displayed along with a link to file a report on github.
Ideally if we could measure how much of the main bundle has been downloaded, we could incorporate that progress into the display, but as far as I can tell there isn't any way to do that.

LeonMontealegre (approved):

PR #781, opened 2021-10-15
Fixed mouse and wire interaction

### LINKED ISSUE #779: Creation and splitting of wires works with the middle and right mouse buttons when it should not.
**Describe the bug**
When a node is clicked and dragged with the middle and right mouse buttons, it creates a wire.  When a created wire is clicked and dragged with the right mouse button, the wire splits. 

**To Reproduce**
1. Place any input component.
2. Attempt to create a node, using the right or middle mouse button instead of left.
**OR**
2. Attempt to split a wire, using the right mouse button instead of left.
3. See that you can successfully create/split a wire with the middle and right mouse buttons.

**Expected behavior**
When clicking and dragging the right mouse button on a node, it should activate selection.
When clicking and dragging the middle mouse button on a node, it should pan.

thedomino1313:
When a node is clicked and dragged with the middle and right mouse buttons, it no longer creates a wire. When a created wire is clicked and dragged with the right mouse button, the wire no longer splits. Both buttons now work as intended.

Fixes #779

### DIFF
--- a/src/app/core/tools/SplitWireTool.ts
+++ b/src/app/core/tools/SplitWireTool.ts
@@ -1,4 +1,4 @@
-import {GRID_SIZE}  from "core/utils/Constants";
+import {GRID_SIZE, LEFT_MOUSE_BUTTON}  from "core/utils/Constants";
 import {V, Vector} from "Vector";
 
 import {Event}       from "core/utils/Events";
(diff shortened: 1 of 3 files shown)


LeonMontealegre:
Can you add a unit test for this? It would be in [this](https://github.com/OpenCircuits/OpenCircuits/tree/master/src/app/tests/digital/utils/tools) folder. Testing should be fairly straightforward and the `FakeInput` class has a `drag` method and you can just pass `RIGHT_MOUSE_BUTTON` and `MIDDLE_MOUSE_BUTTON` to make sure they don't work. To make sure your test works, however, you can temporarily undo your fix and see that the test fails

LeonMontealegre (approved):

PR #782, opened 2021-10-15
XOR gate documentation

vanshreddy:
Added documentation for XOR gate.

### DIFF
--- a/src/app/digital/models/ioobjects/gates/XORGate.ts
+++ b/src/app/digital/models/ioobjects/gates/XORGate.ts
@@ -8,14 +8,20 @@ import {QuadraticCurvePositioner} from "digital/models/ports/positioners/Quadrat
 import {Gate} from "./Gate";
 import {GetQuadraticOffset} from "./ORGate";
 
-
+/***
+ * Xor Gate implementation class. Also know as the inequality detector, returns true when only one of the inputs is a 1.
+ */
 @serializable("XORGate")
 export class XORGate extends Gate {
-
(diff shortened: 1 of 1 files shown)


src/app/digital/models/ioobjects/gates/XORGate.ts
```diff
@@ -24,15 +30,22 @@ export class XORGate extends Gate {
...
+     * Returns a vector which helps position the component's inputs when more inputs are added.
+     */
     // @Override
     public getOffset(): Vector {
         return super.getOffset().add(0, GetQuadraticOffset(this.numInputs()));
     }
-
+    /***
+     * Returns the name of the gate
+     */
```
OwenMcGee, line 42: For instance here, I might write:

/**
\* \@returns string representing the name of the gate
\* /
(if all a function does is return a value, you can probably skip including a description outside of the @returns tag)

OwenMcGee (changes requested):
You aren't using proper JSDoc format for all the functions here, which means that the documentation site will read your comments as lacking. For each function that takes a parameter, for each parameter it takes, there should be a line that looks like: 
`* @param [parameter name] [description of the parameter]`

And for functions that return some value there should be a line that looks like:
`* @returns [description of return value]`

In general, if you type `/**` and then hit enter, most text editors (including VSCode) will then auto construct the outline of the JSDoc comment for you, and you need only add the descriptions.

TGCrystal (approved):

Issue #785, opened 2021-10-15
Panning camera with arrow keys scrolls sidebar (dependent on #786)


RussellABH:
***This is not applicable until #786 is merged***

**Describe the bug**
When you bring in a component and then move the screen with arrow keys, the item nav 
scrolls if you press up or down.

**To Reproduce**
1. Drag in any component from the item nav
2. Press up arrow or down arrow 
3. Notice that item nav scrolls

**Expected behavior**
Either the item nav shouldn't scroll, or the screen shouldn't pan. 
I think the former would be the better solution. 

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Firefox + Chrome (issue not browser specific)

PR #786, opened 2021-10-15
Camera pan with arrow keys

### LINKED ISSUE #707: Allow panning the screen with arrow keys
It would be useful to have the option to pan the canvas using the arrow keys, in addition to the option of using the mouse.

- Could activate whenever the user is not selecting any components on the canvas, and presses a key.
By default, I would expect this to maybe be moving the canvas 1 unit in the given direction (the same amount of movement the arrows in a component's position fields gives), although that would be easier to determine a good amount for once testing the feature in practice
- Holding shift while pressing the arrow keys could give a more granular movement (maybe 0.2 units), although the feature would be fine without this aspect

**Additional context**
For me personally, I find the middle mouse panning awkward on my cheap mouse, and alt is in an awkward position which can cause pain to my wrist to hold down (I have bad tendonitis). I actually find left+right click to be the most comfortable way to pan, despite it being (I believe) an unintentional addition. (Not to say that the current panning options are in any way bad, I just have a bad mouse and even worse arms)

Closely related to (but not dependent upon) #708

RussellABH:
Allows the user to pan the screen with arrow keys. Holding down shift while doing so moves the screen much 
less, allowing precision movement. Works in main canvas and in ICs.

Note that this creates issue #785 

fixes #707


LeonMontealegre:
Also, as noted on your other PR, we should add some unit tests for this,
you'll want to look at the `PanTool.test.ts` file and just add basic tests that make sure that when you press the arrow keys, the camera moves, and then when you press them while holding shift, it moves whatever amount –– I know it seems redundant and potentially useless but these tests are super important so that we can be sure that this feature will always be working even when major parts of the code changes

LeonMontealegre (approved):

PR #787, opened 2021-10-15
More debug options

### LINKED ISSUE #302: More debug options
Render click area for PressableComponents

Render select area for wires

When building the project in dev (non-prod) mode, there should be a way to enable debug modes through some sort of interface that only appears in dev mode

jssunray34:
Closes #302 

I created toggles for each of the four debug options that can be enabled within the settings menu in development mode. This involved creating a state object for debug info.


LeonMontealegre:
Also note that a recently merged PR is going to conflict with yours, it was a refactor of the AutoSaveToggle to show up properly so you're going to have to resolve it locally which you can do by running `git pull origin master` and then fixing it

LeonMontealegre (changes requested):

LeonMontealegre (approved):

PR #790, opened 2021-10-18
Item Nav Drag Preview

### LINKED ISSUE #638: Item preview not appearing when dragging from itemnav
**Describe the bug**
When dragging a component from the itemnav onto the canvas, the item preview visual no longer displays.

**To Reproduce**
1. Go to nightly.opencircuits.io
2. Open the itemnav
3. Drag any component onto the canvas
4. See no preview following cursor

**Expected behavior**
The item preview visual should display following the cursor

 - OS: Windows 10
 - Browser: chrome

**Additional context**
I am almost certain this was introduced by pull #627

### LINKED ISSUE #689: Ability to create Component w/ connected switches/LEDs
It would be very useful to, when dragging a component from the ItemNav, have the placed component come with Switches and LEDs connected to all of its Input/Output ports so it's trivial to test out the component without having to manually hook it up

- Could be triggered by holding something like ALT/OPTION
- Should be considered with #638 and have it preview the connections
- Need a name for this feature, "Auto Place"? "Auto Connect"? "Smart Place"? Please leave any suggestions for naming

A similar feature can be found at `src/app/tests/helpers/Helpers.ts`, currently called "Auto Place"
And the code from there should be extracted out into wherever the feature is placed and then use that in the Helper

LeonMontealegre:
Adds a preview for ItemNav Drag'N'Drop
Also implements the 'Smart Place' feature which can be activated by holding ALT/OPTION while dropping a component from the ItemNav

Fixes #638 
Fixes #689


LeonMontealegre:
@TGCrystal Alright should be fixed, accidentally removed an important LoC right before pushing apparently!

LeonMontealegre:
@OwenMcGee 
1. Yeah this is a good idea
2. (unrelated issue)
3. I cannot reproduce this on mac, also seems like a non-issue for now
4. Yeah, the Bus algorithm needs to be improved, so there's nothing to do about this for now
5. They are as close as possible without overlapping their cull-boxes, you can see that the bottom of the Switch is exactly where the top of the LED is

LeonMontealegre:
Opted for cyclable Smart Place using ALT/Option key. Also allowed users to Press ESCAPE or Right-Click to cancel the placement. Theoretically should also not open the ContextMenu but it's 4 am and it was a little buggy for a bit before I hopefully fixed it. Definitely need to seriously test this @OwenMcGee and see if you can break it

OwenMcGee (approved):
~~Everything seems in working order! The only remaining issues are due to pre-existing problems (namely clock drift) outside the scope of this PR.~~
Wrong PR, I'll replace this with a scathing review soon

TGCrystal (changes requested):
- [x] Currently pressing Alt toggles the connected component place feature on. Pressing it again should toggle it off.
- [ ] Related to the above issue, if you hold the Alt button on that first press, it does rapidly jump between being toggled on and off. If the above is fixed, this isn't that big of an issue and I think would be fine if the rate at which it was toggling off was a bit slower so it was clearer to the user what was happening.
- [x] ~~If you tap a component, then click on the canvas and drag a selection box and release while hovering over an already placed component, then the components you are trying to place do not seem to be placed. I think the best solution would be for the selection box to be disabled while placing a component in that way.~~ Happens on nightly as well, will open as a separate issue
- [x] The preview does not work when using a 2 in 1 touchscreen laptop as a tablet (and likely wouldn't work on mobile either). When dragging and dropping with a finger or pen, the preview image shows up but stays in what appears to be a random location. ~~The tap and place option works if you do the taps on the components and then drag to the canvas.~~ This also happens on nightly, will open as a separate issue.

OwenMcGee (changes requested):
- [ ] Starting with the input section, after pressing alt once to enter smart-place to add LEDS, I would expect pressing alt again to return me to the normal placement mode, but it has no effect. Same with the output components with switches
- [x] Pressing alt once, and _then_ clicking on an input component will make that component start out in smart-place mode, which I wouldn't expect it to do (alt should only do something once a component has already been selected, I feel). I'm fine with the user's smart-place state persisting across components (if I place a switch in smart-place mode, the next time I start placing a button it's a good thing if it starts out assuming to use smart-mode), but the alt key should only affect this state _while_ placing components.
- [x] For components with both input _and_ output ports, I might suggest rather than just cycling between smart place and no smart place, there instead be 4 modes it cycles through:
  1. normal placement
  2. place with switches connected
  3. place with LEDs connected
  4. place with both LEDs and switches connected
- [ ] Holding down alt will rapidly switch smart mode on and off (too rapidly), and if the user releases alt at a time when smart mode is off, pressing alt again will not turn it back on (related to the first issue, I'm sure). Even if the first issue is solved, the rate at which this switches between states should be slowed down (or maybe holding alt should only switch the state once and then be ignored until it is released and pressed again)
- [ ] There's no contingency in place for components (namely, ICs) with output ports on the left or input ports on the right. It should be straightforward enough to define which edge of the IC a port rests on, and position the switches/LEDs appropriately. (I can imagine the logic would get complicated when dealing with input and output on the _same_ side, however). Image of such an IC below: 
image

- [ ] Minor point, but maybe when an IC is tall enough to allow it, switches should prioritize lining up with their ports (as LEDs do) when it wouldn't result in overlap. And similarly, perhaps LEDs could do away with the offset when they wouldn't overlap vertically without it (this one I'm less confident in though, as if the offset is only skipped for a subset of the switches/LEDs it could look strange). Example IC below: 
image

- [ ] I'm having a very difficult time reproducing it, but several times now I've somehow managed to make the placement not work at all (as in, there's a single component selected to be placed, following my cursor, with smart mode off, but when I click on the canvas nothing happens). I'll update this with more information if I can track down what's causing it, but if I can't manage to reproduce it again it may have been some other issue or my computer/mouse acting up, or simply rare enough to be ignored.

TGCrystal (approved):

OwenMcGee:
My first 3 points have been resolved (and the other points were essentially minor suggestions, nothing that should block the merging of the PR), however one new issue has cropped up.

When using smart place with inputs, it takes 2 presses of alt to cycle the feature on and off, but it should only take a single press. This issue does not occur with the output components. When this is fixed, the PR should be good to merge

PR #795, opened 2021-10-18
Resolved Issue #624

### LINKED ISSUE #624: Clicking a node already connected to a wire both starts a new wire and selects the existing wire
**Describe the bug**
If a component has a wire leading out of it and into some other component, clicking that output node (to generate a new wire) will still begin the new wire, but will also select the existing wire.

**To Reproduce**
1. Add a switch and two LEDs to the canvas.
2. Connect the switch to one of the LEDs.
3. Click the output node of the switch
4. The beginning of a new wire will form, but the existing wire will also be selected.

**Expected behavior**
The existing wire should not be selected.

**Screenshots**
(excuse the invisible mouse)
Circuit before clicking
before clicking

Circuit after clicking
after clicking

Expected behavior after clicking
expected behavior

 - OS: Windows 10
 - Browser: chrome

**Additional context**
This does not occur if clicking the very edge of the node, leading me to believe that it only happens when clicking a location that has a wire hidden beneath it. I suspect this is either a result of the recent PR #593, but (far) more likely from issue #621.

mercuryhg31:
Closes #624


src/app/core/tools/InteractionTool.ts
```diff
@@ -10,6 +10,7 @@ import {IOObject} from "core/models";
 import {DefaultTool} from "./DefaultTool";
 
 import {EventHandler} from "./EventHandler";
+import {WiringTool} from "./WiringTool";
```
LeonMontealegre, line 13: So in general it's bad practice to have a tool import some other tool. In theory each tool should be completely independent and not have to explicitly depend on another one.

I'll mention how to fix this below

src/app/core/tools/WiringTool.ts
```diff
@@ -96,6 +125,24 @@ export const WiringTool = (() => {
...
+        },
+
+        /**
+         * To test for special case Node ports
+         * @param info CircuitInfo to be used by findPorts
+         * @returns whether there are visible ports belonging to a Node
+         */
+        nodePorts(info: CircuitInfo): boolean {
+            return findPorts(info).some(o => isNode(o.getParent()));
         }
```
LeonMontealegre, line 146: Since these are the only two functions being used from the `WiringTool` and the `findPorts` function actually doesn't depend on the `WiringTool` at all, we should instead move the `findPorts` function into a different file along with the `visiblePorts` and `nodePorts` functions and then all the files can just import from this new file instead of depending on the `WiringTool`

Specifically, we can move them into `app/core/utils/ComponentUtils.ts`

LeonMontealegre (changes requested):

PR #798, opened 2021-10-19
Bezier Curve Documentation

ezier:


src/app/core/utils/math/BezierCurve.ts
```diff
@@ -1,18 +1,25 @@
-import {Vector,V}  from "./Vector";
-import {Transform} from "./Transform";
-import {Clamp} from "./MathUtils";
-import {serializable} from "serialeazy";
-
+import { Vector, V } from "./Vector";
+import { Transform } from "./Transform";
+import { Clamp } from "./MathUtils";
+import { serializable } from "serialeazy";
```
TGCrystal, line 4: nit: remove the spaces in the import braces, should look like
```ts
import {Clamp} from "./MathUtils";
```

src/app/core/utils/math/BezierCurve.ts
```diff
@@ -1,18 +1,25 @@
...
+
```
TGCrystal, line 5: nit: add extra space after import (should always have two spaces after imports)

src/app/core/utils/math/BezierCurve.ts
```diff
@@ -47,103 +64,121 @@ export class BezierCurve {
...
     public getPos(t: number): Vector {
         return V(this.getX(t), this.getY(t));
     }
 
-    // 1st Derivative
+    // Calculates and returns the 1st derivative of x coord of t
     public getDX(t: number): number {
         const it = 1 - t;
-        return -3*this.p1.x*it*it + 3*this.c1.x*it*(1-3*t) + 3*this.c2.x*t*(2-3*t) + 3*this.p2.x*t*t;
+        return -3 * this.p1.x * it * it + 3 * this.c1.x * it * (1 - 3 * t) + 3 * this.c2.x * t * (2 - 3 * t) + 3 * this.p2.x * t * t;
```
TGCrystal, line 146: Formatting changes like this are probably not necessary

TGCrystal (changes requested):
Make sure the comments you are writing are actual jsdoc comments. jsdoc comments use the `/** */` style of comments instead of `//`
For example, the comment for something like `setP1(v: Vector)` would look something like:
```ts
/**
 * Sets the starting point for the Bezier curve
 *
 * @param v the new starting point
 */
public setP1(v: Vector): void {
```

src/app/core/utils/math/BezierCurve.ts
```diff
@@ -3,16 +3,58 @@ import {Transform} from "./Transform";
...
+    /**
+     * whether the curve's data has been updated
+     */
     private dirty: boolean;
 
+    /**
+     * Initializes bezier curve with the given start point, end point, and intermediate points.
+     * If no point(s) are given, then it will be initialized with a blank Vector.
+     * Bounding Box holds meaningless values by default
+     * @param p1 Initializes start point with given coordinates
```
TGCrystal, line 53: Looking much better now! Just make sure to leave a new line between the end of the description and the first tag, so this section would look like
```
     * Bounding Box holds meaningless values by default
     * 
     * @param p1 Initializes start point with given coordinates
```

src/app/core/utils/math/BezierCurve.ts
```diff
@@ -3,16 +3,59 @@ import {Transform} from "./Transform";
 import {Clamp} from "./MathUtils";
 import {serializable} from "serialeazy";
 
-@serializable("BezierCurve")
+/**
```
TGCrystal, line 6: nit: should have two empty lines between last import and the start of the comment

TGCrystal (changes requested):

TGCrystal (approved):

PR #801, opened 2021-10-19
switched from keycode to key (issue #789)

### LINKED ISSUE #789: Switch from `keyCode` to `key`
[Currently `keyCode` is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) and so we should be safe and make the switch to the `key` constant instead.

This would involve changing all the constants like `SHIFT_KEY` and `A_KEY` to be strings and then updating Input to use `key` instead along with any other event listeners that use `keyCode`.

Kevin-771:
switched key constants to strings and changed event listeners to use key instead of keycode

closes #789


LeonMontealegre:
@Kevin-771 Please fix merge conflicts and also you may need to update this further since more things have been added that use `keyCode/key`

LeonMontealegre (approved):

PR #805, opened 2021-10-22
fixed context menu issue #714

### LINKED ISSUE #714: Context menu gets cut off when cursor is near the bottom of the screen
**Describe the bug**
When the cursor is near the bottom of the screen and the user right clicks, the context menu will simply appear in a position that cuts off the bottom portion of it.

**To Reproduce**
1. Go to https://nightly.opencircuits.io/
2. Move the cursor near the bottom of the screen
3. Right click and observe the context menu being cut off

**Expected behavior**
When the normal behavior would result in the context menu being cut off, I would expect the context menu to either:
 - appear with the bottom left corner at the position of the cursor, rather than the top left
 - appear with only the left edge at the cursor's position, but the vertical position simply as close to the bottom of the screen as it can get (less a slight gap for readability)

**Screenshots**
My cursor is invisible in the screenshots, but pretend it's in the same spot each time (top left corner of the context menu in the first screenshot)

Current behavior:
Screenshot (40)

Expected behavior option 1:
Screenshot (41)

Expected behavior option 2:
Screenshot (42)

 - OS: Windows 10
 - Browser: Chrome

Kevin-771:
made it so that if the context menu were to go offscreen, it would be created on the opposite side of the cursor.

closes #714


src/site/shared/containers/ContextMenu/index.tsx
```diff
@@ -15,6 +15,7 @@ import {useHistory} from "shared/utils/hooks/useHistory";
 
 
 import "./index.scss";
+import { HEADER_HEIGHT } from "shared/utils/Constants";
```
LeonMontealegre, line 18: nit: Remove spaces inside `{}`

LeonMontealegre: nit: Move constants to be the top import under import from "react"

src/site/shared/containers/ContextMenu/index.tsx
```diff
@@ -24,6 +25,8 @@ function isClipboardSupported(type: "read" | "write"): boolean {
 }
 
 
+const CONTEXT_MENU_OFFSET = 4;
```
LeonMontealegre, line 28: What does this represent? Add a comment?

LeonMontealegre (changes requested):
This doesn't appear to fix the issue as far as I can tell, it still appears to go off screen

LeonMontealegre (approved):

PR #819, opened 2021-10-29
Power Wires During Creation

### LINKED ISSUE #704: Powered wires don't appear blue during their creation
**Describe the bug**
Powered wires always appear blue, however when creating a new wire from an output port that is sending power, the wire will not appear blue until after it has been connected to an input port.

**To Reproduce**
1. Place a constant high and an LED on the canvas
2. Create a new wire originating from the constant high
3. Observe that the wire is white, not blue
4. Connect the wire to the LED and observe its color snap to blue

**Expected behavior**
I would expect the wire to be blue even during creation

**Screenshots**
Current behavior (wire is not blue):
Screenshot (36)
Wire becoming blue once connected to an input port:
Screenshot (37)

 - OS: Windows 10
 - Browser: chrome

**Additional context**
The case of wires created from input ports and _then_ connected to powered output ports is already working as it should, and doesn't need any changes.

riedee:
Fixes #704 

If a wire is created and its parent is active (e.g. ConstantHigh), the wire will be powered. If the parent is inactive (e.g. ConstantLow), the wire will not be powered. Also added code in Clock.ts so each tick will update the wire in WiringTool (if necessary).


OwenMcGee (changes requested):
The issue of the wire not updating as a clock switches state still persists, in the cases where any component (other than the clock itself) changes its output state due to a clock changing state, as in the below example:

Animation2

In fact, it need not be a clock, just any component whose output state is changing for any reason, as in the following example of a pulse cycling through a closed loop:

Animation3

src/app/digital/models/ioobjects/inputs/Clock.ts
```diff
@@ -33,6 +35,13 @@ export class Clock extends DigitalComponent {
     public activate(bool: boolean): void {
         super.activate(bool);
         this.isOn = bool;
+
+        // If WiringTool is active (and its parent is a clock), activate its wire as well
+        if (WiringTool.getWire() !== undefined && WiringTool.getWire().getP1Component().getName() == "Clock") {
+            const w = WiringTool.getWire() as DigitalWire;
+            w.activate(bool);
+        }
+
```
LeonMontealegre, line 44: So as I mentioned in class, the idea of importing something like the `WiringTool` into something like the `Clock` tends to be a bad idea. (and also again don't feel bad, it's super awesome that you were able to solve the issue regardless and I never mentioned that this is "bad" practice).

The reason I would say this is "bad" is because the Clock is an isolated component that really shouldn't know anything about the WiringTool's existence, and I can imagine a scenario where the Clock exists despite the WiringTool not existing like inside an IC or something like that.

But anyways –– I think the solution here is to have the wire be drawn as On/Off completely depending on whether the wire's `getInput().getIsOn()` and so the solution would be located probably in `src/app/digital/rendering/ToolRenderer` where you should be able to find the rendering for the `WiringTool`

LeonMontealegre (changes requested):

src/app/digital/models/ioobjects/inputs/Clock.ts
riedee: Yeah, I realized when you mentioned that in class the other day that my solution was definitely not the best.
I was working on it a bit more and I thought I found another solution, but it was causing errors when I was running the yarn tests, so I tried yours and was able to get everything working with only one line change in WireRenderer (plus what I had done in DigitalWire).

I'm just confused about one thing, I noticed it when I was testing my other solution as well:
opencirc
opencirc2
Is it expected that not gates, nands, nors, etc. power wires without a source?
Besides that everything seems good, I just wanted to make sure it's expected before I commit.

LeonMontealegre: Awesome! And yeah that should be totally fine for NOTGate's and such

src/app/digital/rendering/ioobjects/WireRenderer.ts
```diff
@@ -45,7 +45,7 @@ export const WireRenderer = (() => {
             ));
                 
             // @TODO move to function for getting color based on being selection/on/off
-            const color = (wire.getIsOn() ? DEFAULT_ON_COLOR : (selected ? selectedColor : wire.getColor()));
+            const color = (wire.getInput().getIsOn() ? DEFAULT_ON_COLOR : (selected ? selectedColor : wire.getColor()));
```
LeonMontealegre, line 48: What happens if you try click on an LED's Input Port to connect the wire?
I have a feeling this might break? If not then lmk and we should be good to merge!

riedee: You're right, it did, I didn't even think of that.
I made a quick change that I just tested and I believe will fix the issue, so let me know if it looks okay.

src/app/digital/models/DigitalWire.ts
```diff
@@ -23,7 +23,12 @@ export class DigitalWire extends Wire {
 
         this.p1 = input;
         this.p2 = output;
-        this.isOn = false;
+        if (input !== undefined) {
+            this.isOn = input.getIsOn();
+        }
+        else {
+            this.isOn = undefined;
+        }
```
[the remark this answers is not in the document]
riedee: I thought it still would be but I'm testing it now and no, it doesn't seem to be. 
I'll commit once more without this change once I've tested it a bit more.

src/app/digital/rendering/ioobjects/WireRenderer.ts
```diff
@@ -45,7 +45,13 @@ export const WireRenderer = (() => {
...
                 
             // @TODO move to function for getting color based on being selection/on/off
-            const color = (wire.getIsOn() ? DEFAULT_ON_COLOR : (selected ? selectedColor : wire.getColor()));
+            let color;
+            if (wire.getInput() !== undefined) {
+                color = (wire.getInput().getIsOn() ? DEFAULT_ON_COLOR : (selected ? selectedColor : wire.getColor()));
+            }
+            else {
+                color = selected ? selectedColor : wire.getColor();
+            }
```
LeonMontealegre, line 54: I think this can be condensed back down to what you had using a cool new-ish TypeScript feature called "optional chaining". Basically if you put a `?` before the `.getIsOn()` it will check if `wire.getInput()` is undefined and if it is, will break out of the statement and return `false`, but then if `wire.getInput()` is defined it will be able to call `.getIsOn()` and then work as normal ([here is some documentation on it](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining))

src/app/digital/models/DigitalWire.ts
```diff
@@ -23,7 +23,7 @@ export class DigitalWire extends Wire {
 
         this.p1 = input;
         this.p2 = output;
-        this.isOn = false;
+        this.isOn = false;
```
LeonMontealegre, line 26: nit: remove the extra space at the end here

src/app/digital/models/DigitalWire.ts
```diff
@@ -74,4 +74,4 @@ export class DigitalWire extends Wire {
     public getIsOn(): boolean {
         return this.isOn;
     }
-}
+}
```
LeonMontealegre, line 77: nit: put the blank last line back so that this file basically just becomes un-edited

LeonMontealegre (changes requested):

LeonMontealegre (approved):

PR #821, opened 2021-10-29
fixed wonky wire connections (issue 610)

### LINKED ISSUE #610: Segment display sometimes connects wires to wrong ports
When attaching a wire to a port of a segment display, it will often attach to the port above the intended one, even if the cursor is directly over the center of the port. After connecting to the incorrect port, if the wire is again dragged to the same port, it will then correctly connect.

I suspect this happens because the port hitboxes are oversized (which by itself isn't an issue), but when selecting an overlapping region no check is done to see which port the cursor is actually closest to. Currently the precedence appears to be to prioritize ports closer to the top of the segment display (that is, top during default orientation. When rotated 180 degrees, the now bottom port is still the highest precedence).

Kevin-771:
fixed so that the wires will connect to the closest port.

fixes #610


Kevin-771:
working to make ports on top have priority over ports underneath regardless of distance.

LeonMontealegre (approved):

PR #828, opened 2021-10-30
Documentation for Shapes

jssunray34:
Wrote documentation for shapes classes.

### DIFF
--- a/src/app/core/rendering/shapes/ArcCircle.ts
+++ b/src/app/core/rendering/shapes/ArcCircle.ts
@@ -11,6 +11,7 @@ export class ArcCircle extends Circle {
 
     /**
      * Constructor for ArcCircle
+     * 
      * @param pos the position
      * @param radius the radius
      * @param a0 the starting angle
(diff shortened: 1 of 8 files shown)


src/app/core/rendering/shapes/ArcCircle.ts
```diff
@@ -2,16 +2,30 @@ import {Vector} from "Vector";
...
+/**
+ * A representation of an ArcCircle shape.
+ */
 export class ArcCircle extends Circle {
     protected a0: number = 0;
     protected a1: number = 2*Math.PI;
 
+    /**
+     * Constructor for ArcCircle
+     * @param pos the position
```
TGCrystal, line 14: nit: Leave an empty line between the end of the description and the start of the tags like so:
```
 * Constructor for ArcCircle
 *
 * @param pos the position
```

src/app/core/rendering/shapes/ArcCircle.ts
```diff
@@ -2,16 +2,30 @@ import {Vector} from "Vector";
 
 import {Circle} from "./Circle";
```
TGCrystal, line 4: Add another space between last import and the start of the comment

TGCrystal (approved):

Issue #613, opened 2021-03-09
Duplicated components staying powered with no power source


Xystem4:
**Description**
When duplicating a component (or group of components), but _not_ the power source of the component(s), the copy will behave as if it is still being powered. This occurs when duplicating with ctrl+d, dragging and pressing space, and (on opencircuits.io where copy+paste still works) when using ctrl+c ctrl+v.

Updating a component stuck this way does not fix it. Moving it, changing where its output connects to, changing the number of inputs it takes, none stop it from staying powered. Only connecting a new power source and then removing it fixes it (or connecting an unpowered wire to the input port).

Note: this seems to affect each input port individually. Connecting an unpowered wire to an input port that thinks it is being powered will fix _that_ port, but any others will remain unchanged.

**To reproduce**
create a power source, and then one more more components taking power from it.
Select any amount of components receiving power to one or more input port, _without_ selecting the power source.
Duplicate the selected nodes

**Screenshot**
The entire top circuit was selected, except for the power source. It was then duplicated to form the other two circuits below. The bottom circuit then had one buffer deleted, updating the right half.
![duplicate bug](https://user-images.githubusercontent.com/65842664/110548868-3c342c00-80ff-11eb-86cf-dfe4d417f61f.png)

TGCrystal:
Likely due to the same underlying issue as #468

PR #839, opened 2021-11-03
Oscilloscope

### LINKED ISSUE #500: Oscilloscope

LeonMontealegre:
Added an Oscilloscope component

https://user-images.githubusercontent.com/29997057/140170609-a4cd3aa3-9667-4733-9cdc-00c9e9039a1d.mov

Todo:
- [x] Better icon on ItemNav
- [x] Fix name-wrapping on ItemNav somehow
- [x] Abstract out shared methods/variables b/t Clock/Oscilloscope => `TimedComponent`
- [x] Mutable width/height?
- [x] Maybe multiple inputs => multiple graphs?
  - [x] ~~Color setting for each input~~
  - [x] ~~Port label~~
  - [x] ~~Visible legend?~~

Closes #500


TGCrystal:
On the topic of the name wrapping in item nav, I think there are a few ways that can be addressed. The Oscilloscope icon could take up two columns (and it is a larger component so that could sort of make sense) and then the text wouldn't have to wrap. Alternatively, the name could be hyphenated like so: 
image
Alternatively, the name itself can be shortened to o-scope or something like that.

As for bugs:

- [ ] The Delay changes in increments of 100 when using the buttons in the text box, so that means decrementing from the starting 100 goes to 50 and incrementing skips back up past 100 to 150 and now without specifying an exact delay, it can't be what it originally was. This isn't that big of an issue, but imo the delay should just change in increments of 50 instead.
- [ ] If you duplicate the oscilloscope when it is "on" the duplicated one will remain counted as on. This may just be another extension of #613 
- [ ] Viewing an oscilloscope connected to a switch inside an IC only updates the state when you are looking at it, so if you let it sit closed for a few seconds then view the IC it looks like this 
image

OwenMcGee:
- [ ] Building off what @TGCrystal said about the delay increments, it appears that it actually doesn't consistently move by 100, and simply gets stopped when going down from 100, it will _always_ go to a multiple of 50, and if it isn't at one it will show a little message box when the cursor is over the input field, saying "Please enter a valid value. The two nearest valid values are [50] and [150]". I agree that it should behave as other components do, and the arrows should simply move up or down 50 units, regardless of if the new value is a multiple of 50 or not. Gif displaying the behavior (couldn't get it in just a screenshot):
Animation2
It's worth noting that these values that aren't multiples of 50 _do_ have an effect on the delay time, and properly work and display differently than simply one of the nearest two multiples of 50.

- [ ] when increasing the samples, the oscilloscope on the canvas immediately updates to display according to the new value. However, when the oscilloscope is deselected, it snaps back to only showing the left half (although not the same left half that was originally displayed when first switching). Gif displaying this (the oscilloscope on the right is just so the selection popup is out of the way):
Animation

- [ ] Also, a strange issue where oscilloscopes that should display exactly the same have some minor differences between them. In the gif below, I have two oscilloscopes with identical (default) delay and samples, connected to the same clock, each the same distance from the clock. When I pause the oscilloscopes however and put them on top of one another, switching which is on focus (clicking puts the bottommost component on top, still haven't fixed that) reveals there is clearly a slight difference between the two. It's somewhat visible (I think) while they're unpaused as well, the two seem somewhat "jittery" compared to one another as the line moves.
Animation3

All in all, not much of consequence wrong with it. If multiple inputs gets added, maybe take the color from the color of the wire attached to that input? (with the option to change it, still). Since naming the ports is an option, if a legend were added the lines could be labelled according to the port title (or maybe the name of whatever component the wire into the oscilloscope came from). I've never used oscilloscopes too much myself so I'm unsure if there's anything glaring missing from this, but it seems good to me.

LeonMontealegre:
>  Viewing an oscilloscope connected to a switch inside an IC only updates the state when you are looking at it, so if you let it sit closed for a few seconds then view the IC it looks like this

Yeah probably should just not allow oscilloscopes in ICs

> Weird behavior when going back to that tab from a different tab (the circuit here was not changed)

Yeah, not sure what the best way to resolve this is -- they should probably just pause when the user isn't focused? I don't think it's possible to allow the page to continue to propagate while out of focus or minimized or whatever

> when increasing the samples, the oscilloscope on the canvas immediately updates to display according to the new value. However, when the oscilloscope is deselected, it snaps back to only showing the left half (although not the same left half that was originally displayed when first switching). Gif displaying this (the oscilloscope on the right is just so the selection popup is out of the way):

I think I figured out what the problem here is: The selection popup creates temporary actions to change the # of samples and then when it creates the final action, it undoes the temporary one which then cuts off the extra values temporarily

>  Also, a strange issue where oscilloscopes that should display exactly the same have some minor differences between them. In the gif below, I have two oscilloscopes with identical (default) delay and samples, connected to the same clock, each the same distance from the clock. When I pause the oscilloscopes however and put them on top of one another, switching which is on focus (clicking puts the bottommost component on top, still haven't fixed that) reveals there is clearly a slight difference between the two. It's somewhat visible (I think) while they're unpaused as well, the two seem somewhat "jittery" compared to one another as the line moves.

This is weird, how did you sync them in the first place? Pause them, clear them, then resume them at the same time?

TGCrystal:
> Yeah probably should just not allow oscilloscopes in ICs

I agree, but I think that if that is the case then there should be some way of letting the user know that the reason the IC they are trying to create isn't being created is because it has an oscilloscope in it. Otherwise I can almost guarantee there would be an issue called something like "Can't create IC with oscilloscope" opened within a year of merging this.
Alternatively, I just thought of a different idea: would it just be possible for the oscilloscope view in the IC to just be cleared every time it is opened? That way you don't have to worry about undefined behaviors that happen while it is closed, but you could still use them in ICs.



> Yeah, not sure what the best way to resolve this is -- they should probably just pause when the page isn't focused? I don't think it's possible to allow the page to continue to propagate while out of focus or minimized or whatever

I agree, that makes the most sense. Even if it was able to propagate while the page was out of focus, that would just be spending unnecessary system resources.

src/app/digital/actions/FrequencyChangeAction.ts
```diff
@@ -0,0 +1,32 @@
...
+import {Oscilloscope} from "digital/models/ioobjects/outputs/Oscilloscope";
+
+
+export class FrequencyChangeAction implements Action {
+    private component: Clock | Oscilloscope;
+
+    private initialFreq: number;
+    private targetFreq: number;
+
+    public constructor(component: Clock | Oscilloscope, targetFreq: number) {
```
TGCrystal, line 13: Should there be a `type FrequencyComponent = Clock | Oscilloscope` or are the amount of uses for that too few?

LeonMontealegre: I'm probably gonna make a `TimedComponent` class that Clock and Oscilloscope will inherit from since they have quite the few similarities

LeonMontealegre:
@OwenMcGee Alright, all the given issues should be fixed - please try and break it :D

(also yes I realize that the clocks drift which causes the multi-graph to drift and should be a separate issue)

OwenMcGee:
- [ ] The arrows now change the delay amount by 50 rather than 100, which is good, but the change also applied to normal clocks (I believe you consolidated some code that would apply to various components dealing with timing), where the old value of 100 seems to make more sense. That's up for debate though, maybe I'm just too used to it being 100, I can see the benefits of having them be consistent
- [ ] The "delay" and "samples" input fields still somewhat treat valid inputs as invalid. Anything in Delay not a multiple of 50, or anything in Samples not a multiple of 10, will when hovered over give the "please enter a valid value..." message, and when using the up/down arrows will jump to the nearest multiple of 10/50 rather than increasing/decreasing by exactly 10/50. However these values are _accepted_ (unlike something like a fractional input, which is automatically changed to an integer once the input field is no longer in focus), and do visibly affect the oscilloscope, so I don't think they should be treated as invalid (or if they should be, they should be consistently treated as invalid, and not allowed as inputs). Upon further inspection, this behavior was already present in clocks before this branch, but I still believe it to be an error (and should not be carried over to oscilloscopes).
- [ ] Now, the big one. If you make an oscilloscope, and then pause it, and then duplicate it (copy paste, ctrl+d, translate+space, etc.), the copy will not be paused (but its selection popup will still have a "resume" option, as if it was). Clearing this copy will then result in an error and some bad visual issues, shown below (first I'm zooming, then I'm panning):smaller I at some point managed to make this just loop forever increasing the error count, but I'm still trying to track down exactly what caused that, I haven't been able to reproduce it completely

OwenMcGee (changes requested):
- [ ] ~~One potential issue is that some clock speeds are too quick to be properly picked up by the Oscilloscope. In the below screenshot, I have synced clocks with delays of 50, 100, 200, 400, and so on, and all are properly displayed by the oscilloscope with the exception of the top clock, with a delay of 50. It appears that the clock is not being detected properly, not just that it isn't being drawn properly~~
image

- [ ] For oscilloscopes that are less wide, sometimes some strangeness comes up for faster clocks (but not too fast to be detected) in drawing them, as with the second and third clocks in the below example. This :
image

- [ ] I believe the pause option for clocks is something you added in this PR. Pausing works properly, but every time a clock is _resumed_ it switches state, which it shouldn't do. (it is nice being able to simultaneously pause clocks and the oscilloscopes they are connected to, though)

LeonMontealegre:
The last 2 points can be addressed when fixing the clock-drift

OwenMcGee (approved):
Everything seems in working order! The only remaining issues are due to pre-existing problems (namely clock drift) outside the scope of this PR.

PR #856, opened 2021-11-05
Prevent zoomin gesture

### LINKED ISSUE #745: Add touch-pad gesture recognition
When you use touchpad on laptop to zoom in the circuit, this will both zoom in the circuit and brower, which is inconvenient.

BaiTueeChuan:
This is to prvent on all browers' default zoomin gesture.

Closes #745


BaiTueeChuan:
This is #745

src/app/core/utils/Input.ts
```diff
@@ -61,7 +60,6 @@ export class Input {
             [[S_KEY], [CONTROL_KEY, COMMAND_KEY, META_KEY]],
             [[D_KEY], [CONTROL_KEY, COMMAND_KEY, META_KEY]],
             [[BACKSPACE_KEY]],
-            [[OPTION_KEY]], // Needed because Alt on Chrome on Windows/Linux causes page to lose focus
```
TGCrystal, line 64: Was this supposed to be removed? I can't see why this would conflict with what this pull request is trying to do.

src/app/core/utils/Input.ts
```diff
@@ -189,9 +201,6 @@ export class Input {
     public addListener(listener: Listener): void {
         this.listeners.push(listener);
     }
-    public removeListener(listener: Listener): void {
-        this.listeners.splice(this.listeners.indexOf(listener), 1);
-    }
```
TGCrystal, line 194: I'm not sure if this function was supposed to be deleted, when trying to run this branch locally I get an error because
```
Property 'removeListener' does not exist on type 'Input'.
```

src/app/core/utils/Input.ts
```diff
@@ -204,11 +213,6 @@ export class Input {
     public isShiftKeyDown(): boolean {
         return this.isKeyDown(SHIFT_KEY);
     }
-
-    public isEscKeyDown(): boolean {
-        return this.isKeyDown(ESC_KEY);
-    }
-
```
TGCrystal, line 211: Similarly, I'm not entirely sure why this deletion was made. Was it accidental?

TGCrystal (changes requested):

src/app/core/utils/Input.ts
BaiTueeChuan: This don't need to be removed. I am not sure why this is removed.

src/app/core/utils/Input.ts
BaiTueeChuan: The main change of this file should only be the function I added and one change I added to touchManager. I am not sure why the rest of changes appear. I am gonna update the code, redo this change and try to make sure this function works without deleting others.

LeonMontealegre (approved):

PR #860, opened 2021-11-06
False power source upon duplication fix

### LINKED ISSUE #468: ICs don't propagate signals immediately when connected

### LINKED ISSUE #613: Duplicated components staying powered with no power source

### LINKED ISSUE #836: Input ports staying powered after disconnected from quickly oscillating wires
**Describe the bug**
Sometimes when removing a wire that was powered, but quickly switching states between powered and unpowered, the input port it was connected to will behave as if it is still receiving power, even when the wire is deleted.

**To Reproduce**
1. Create a closed cycle with a short pulse going through it
2. Delete one wire, and observe the rest of the cycle staying powered (may take a few tries)

**Expected behavior**
I would expect the input port to recognize it is not being powered.

**Screenshots**
My best attempt at making a gif of the layout I used to produce this (the oscillations are occurring much faster on the site than the gif was able to pick up/display):
Animation5

It's also possible to do with a normal clock (although much harder to do), I _believe_ when the wire is deleted just as it's switching from unpowered to powered. I'll just put a screenshot for the outcome of this one:
Screenshot (89)

 - OS: Windows 10
 - Browser: Chrome

**Additional context**
Almost certainly related to issues #613 and #468

mercuryhg31:
Resolves #613, resolves #836


mercuryhg31:
@LeonMontealegre  We discussed making a digital CopyGroupAction, buuuut that's going to mean making a digital version of everything that uses it, and everything that uses that, and so on and so forth, please advise. Some things like CopyGroupAction I can just extend and make slight digital alterations, but I don't think I can not copy whole blocks of code for the things that use it, and that seems like a bad idea.

mercuryhg31:
Noting this super weird failed test case down to fix:

https://user-images.githubusercontent.com/26150304/141033133-e5397e9d-dac1-4cb2-9f5c-422624d66557.mp4

LeonMontealegre:
In terms of that weird bug - I lied to you! Can't just do `getInputPorts()` because the `Mux` and has SelectPorts which aren't exactly input ports (but maybe they should be but that's besides the point), so will have to loop through `getPorts().filter(p => p instanceof InputPort) as InputPort[]` probably

LeonMontealegre (approved):

PR #900, opened 2021-11-19
Switched keycode to key V2.0

### LINKED ISSUE #789: Switch from `keyCode` to `key`

Kevin-771:
changed keycode numbers to key strings

fixes #789


Kevin-771:
so far it works for windows, i don't have a mac to test the option key

Kevin-771:
so far it works for windows, i don't have a mac to test the option key

PR #904, opened 2021-11-19
Translate with arrow keys

### LINKED ISSUE #708: Allow moving components with arrow keys
It would be useful to be able to move selected components using the arrow keys, in addition to being able to drag them with the mouse.

- If the user has a component selected, pressing an arrow key would move the component 1 unit (exact numbers can be tweaked if needed) in the given direction
- If multiple components are selected, each will be moved 1 unit in the given direction, relative to its own starting location
- holding shift while pressing the arrow key could allow for more granular movement, perhaps 0.1 or 0.2 units (similar to how moving objects/images in powerpoint/google slides works, for example)

**Additional context**
Closely related to (but not dependent upon) #707

RussellABH:
Made it possible to translate the selected components using the arrow keys. 

- Uses the translate tool instead of modifying the positions, so relative component placement is maintained. 
- Holding shift translates a smaller amount 
- Added new constants for the amount that the components move
  - Right now the smaller amount is 11 so the component actually moves, less than 11 will be within snapping distance so the component would not move.

Fixes #708


RussellABH:
Sorry, looks like I forgot to run the tests. I'll look into the test that failed.

src/app/core/tools/PanTool.ts
```diff
@@ -8,17 +8,21 @@ import {CircuitInfo} from "core/utils/CircuitInfo";
 
 import {Tool}        from "core/tools/Tool";
 
+import {Component}   from "core/models";
+
```
LeonMontealegre, line 12: nit: two lines between import and first LoC

src/app/core/tools/PanTool.ts
```diff
@@ -8,17 +8,21 @@ import {CircuitInfo} from "core/utils/CircuitInfo";
...
-        shouldActivate(event: Event, {input}: CircuitInfo): boolean {
+        shouldActivate(event: Event, {input, selections}: CircuitInfo): boolean {
             // Activate if the user just pressed the "option key"
             //  or if the user began dragging with either 2 fingers
             //                           or the middle mouse button
-            //  or if the user pressed one of of the arrow keys
-            return (event.type === "keydown"   && ((event.key === OPTION_KEY) ||  
+            //  or if the user pressed one of of the arrow keys while no components are selected
+            return ((event.type === "keydown" && 
+                  !((selections.all((c) => c instanceof Component) && selections.get().length > 0)) &&
```
LeonMontealegre, line 22: Wait why do we need all of the selections to be components for the **Pan** tool?

src/app/core/tools/TranslateTool.ts
```diff
@@ -31,26 +32,37 @@ export const TranslateTool: Tool = (() => {
...
+            (event.type === "keyup" && event.key == activatedButton &&
+                                       (event.key == ARROW_LEFT || event.key == ARROW_RIGHT || 
+                                        event.key == ARROW_UP   || event.key == ARROW_DOWN ));
         },
 
 
         onActivate(event: Event, info: CircuitInfo): void {
             const {camera, input, selections, currentlyPressedObject, designer} = info;
 
+            console.log("i have been activated");
```
LeonMontealegre, line 53: nit: Remove log

src/app/core/tools/TranslateTool.ts
```diff
@@ -82,6 +94,8 @@ export const TranslateTool: Tool = (() => {
                 //  So instead mousemove is used and whether or not left mouse is still pressed is
                 //  handled within the activation and deactivation of this tool.
                 case "mousemove":
+                    if (activatedButton != LEFT_MOUSE_BUTTON) break;
```
LeonMontealegre, line 97: nit: `!==`

src/app/core/tools/TranslateTool.ts
```diff
@@ -95,10 +109,9 @@ export const TranslateTool: Tool = (() => {
                         curPositions;
 
                     // Execute translate but don't save to group
-                    //  action since we do that onDeactivate
                     new TranslateAction(components, initalPositions, newPositions).execute();
 
-                    return true;
+                    return true;
```
LeonMontealegre, line 114: nit: Remove extra space at end of line

src/app/core/tools/TranslateTool.ts
```diff
@@ -107,6 +120,30 @@ export const TranslateTool: Tool = (() => {
                         return true;
                     }
                     break;
+                
+                case "keydown":
+                    if (activatedButton == LEFT_MOUSE_BUTTON) break;
```
LeonMontealegre, line 125: nit: use `===`

src/app/core/tools/TranslateTool.ts
```diff
@@ -107,6 +120,30 @@ export const TranslateTool: Tool = (() => {
...
+
+                    //Translate with the arrow keys
```
LeonMontealegre, line 127: nit: Put space after `//`

LeonMontealegre (changes requested):

src/app/core/tools/PanTool.ts
TGCrystal: To my understanding, this checks to make sure that arrow keys don't pan when components are selected (since it should be translating those components instead)

LeonMontealegre: This should be done implicitly by the TranslateTool being before the PanTool in the list of tools (https://docs.google.com/drawings/d/11e6BiEQTNM3IdJawx4hk_HOe3uSZYisY_ICu5q6E8cY/edit?usp=sharing)

TGCrystal: That doesn't quite seem to do it. Removing the `!((selections.all((c) => c instanceof Component) && selections.get().length > 0))` condition causes the arrow keys to not translate at all, only pan. Removing just the `(selections.all((c) => c instanceof Component)` part of the condition allows for translating, but if just ports are selected, for example, then panning is blocked even though no translating is happening.

LeonMontealegre: You will have to put TranslateTool before the PanTool in the list of tools for it to work lol

LeonMontealegre (approved):

PR #913, opened 2021-11-23
Add ability to paste at mouse location with context menu

### LINKED ISSUE #747: When pasting with mouse, put new components at the location of the cursor
Currently when pasting (mouse or keyboard), the new component(s) will always be placed slightly down and to the right of the location of the original component(s) (regardless of if those original components have moved).

However, when pasting with the right-click menu, it would make more sense for the component(s) to be pasted at the location of the cursor instead, which is likely closer to the user's actual desired location for the new components.

mercuryhg31:
Closes #747 
Also added some docs


src/site/pages/digital/src/utils/DigitalPaste.ts
```diff
@@ -59,16 +64,41 @@ function TransferNewICData(objs: IOObject[], designer: DigitalCircuitDesigner):
...
+ * Maps positions of pasted components based around intially selected item
+ * @param comps Components in clipboard
+ * @param info Circuit info from which to pull camera and input
+ * @returns Target positions for pasting at mouse pos
+ */
+function shiftAllObj(comps: Component[], info: DigitalCircuitInfo): Vector[] {
+    const {camera, input} = info;
+    let worldMousePos = camera.getWorldPos(input.getMousePos());
+    let shift = comps[0].getPos().sub(worldMousePos);
+    return comps.map(o => o.getPos().sub(shift))
```
LeonMontealegre, line 77: nit: in my head it makes more sense to have `shift = worldMousePos.sub(comps[0].getPos())` and then do `o.getPos().add(shift)`

src/site/pages/digital/src/utils/DigitalPaste.ts
```diff
@@ -59,16 +64,41 @@ function TransferNewICData(objs: IOObject[], designer: DigitalCircuitDesigner):
...
+        const {history, designer, selections, renderer} = info;
         const objs = Deserialize<IOObject[]>(data);
 
         const newICDataAction = TransferNewICData(objs, designer);
 
         // Get all components
         const comps = objs.filter(o => o instanceof Component) as Component[];
 
+        // Determine target positions for pasted components
+        let targetPositions = mouse ?
```
LeonMontealegre, line 98: nit: use `const`

LeonMontealegre (changes requested):
Awesome PR! Almost perfect besides a few tiny nits! Highly appreciate the JSDoc's comments as well

OwenMcGee:
I'm not sure if this is intended or not, but I'll just note that the components get pasted based on where your cursor was last placed on the canvas, _not_ where it was when the user right clicked. This means that if the user right clicks, then moves their cursor outside the right click menu and then re-enters the right click menu from a different corner (different from the typical top-left corner, that is), the components will be pasted in a different location.

The wonkiest you can get with this is right clicking somewhere, then moving the cursor to the complete opposite side of the canvas, then alt-tab away, then move the cursor to where the right click menu should be on the screen, and then alt-tab back, then click paste, to get the components pasted far away from where the right click menu was.

Not really a big deal, the alternatives I see would be always pasting exactly where the user clicked the right mouse button, or always pasting at the top left corner of the right click menu (not necessarily the same, consider if the user zooms). Just an interesting quirk to note.

Aside from that, things seem to be working perfectly well!
seal of approval

mercuryhg31:
That wasn't intentional and I think it'd be confusing behavior for a user not thinking about their mouse position. That wonky case isn't super far out of the realm of possibility either so I've fixed it to stick to the context menu position.

I did notice something odd about the selection popup while testing this out though and I'm not sure why this is happening:

https://user-images.githubusercontent.com/26150304/144547372-a1a7dbb6-22e3-4a5c-bbd2-aea4a0f05303.mp4

Sometimes the selection popup will be located somewhere between the original group of components and the duplicate. Above this only happens at 0:10, 0:20, and 0:25, the other times, it works just fine. Not sure if this is related to pasting at mouse location or a quirk of the way selection popup works.

OwenMcGee:
The components are now consistently pasting at the exact location I right click, so that's all good now.

As for your other issue, it's due to an existing problem.

If you look at your video, you can see that every time the strange location of the selectionpopup happens, it's actually the _same_ strange location. That being the location of the original group of components you copied. Since paste with mouse has been working, there's been the slight bug that whenever you paste, for a split second the selectionpopup first acts as if the original components are what is selected (although to my knowledge they aren't ever highlighted green again, even for a split second), meaning it shows their information and appears just to the right of those components.

On nightly this is difficult to notice, since the components are only ever pasted just next to the original group, and it fixes itself after a few frames. You can prolong the time it takes by pasting in such a way that after clicking paste in the context menu, your cursor will be somewhere over the selectionpopup of the original components. When doing this, it takes longer (about a second) for the selectionpopup to fix itself (easier to see when at least one of the coordinate text boxes has an actual number in it). (actually on a second look, after about a second the contents of the popup (coordinates) fix themselves, but the location doesn't get fixed until the cursor is moved off the popup, as in your video). This is an existing bug that's not caused by your PR, and so not really your problem.

What _might_ be your problem though, is the reintroduction (I think) of an older issue, that being that if you don't move the cursor at all after clicking paste, the selectionpopup does not update to be next to the newly pasted components (though they are properly highlighted green, with the red rotation circle there). Once a key is pressed or the cursor is moved, the popup updates. (this is far easier to replicate with a trackpad than with a mouse).

This issue definitely existed in the past (I have it written down in my notes for bugs to document), but I can't seem to replicate it on nightly right now (but maybe I'm just not trying hard enough). So, that one might be your responsibility.

mercuryhg31:
Okay, it does appear to only not update the selection popup location when my mouse or the context menu is on the popup, which is the same as nightly so I'll leave it be, it's just a much more noticeable bug with pasting at position.

LeonMontealegre (approved):

PR #916, opened 2021-11-24
Image Exporter

### LINKED ISSUE #912: Circuit Expression Generator toggles don't have correct cursor on hover
The toggles for each option w/in the Circuit Expression Generator popup don't turn to `cursor: pointer` when hovered

LeonMontealegre:
Added an "image exporter" which is a popup that allows customization of the exported image, replaces the individual Save As PNG/JPEG/PDF menu items

Fixes #912


TGCrystal (changes requested):
- [ ] There should be a button to reset the Background color to the default, it looks like there would be room to fit it in on the right here 
image

- [ ] `Background color` should be `Background Color` (considering `File Type` capitalizes the first letter of each word)
- [ ] There should be a button to fit the entire circuit inside of the current view.
- [ ] Disabling the Background makes it look kind of like the wires (and anything white) are invisible. Maybe when background is off, the grid look used to indicate transparency can be used as a background in the preview. This can also help show the size of what is getting saved when Background is disabled.
- [ ] Export as PDF not looking like it should on iOS (PNG and JPEG do look like the preview)
image image

LeonMontealegre:
>  Export as PDF not looking like it should on iOS (PNG and JPEG do look like the preview)

PDF seems even worse in nightly so I would call this an improvement – also this is such a small use-case (like who the hell needs a PDF form of their circuit on their phones) so I'm gonna say it's out of scope


> There should be a button to fit the entire circuit inside of the current view.

Definitely allowing them to press "F" is a good plan - do you think there should be a physical button too?


> Disabling the Background makes it look kind of like the wires (and anything white) are invi...

You're talking about this pattern? Yeah I couldn't find a trivial way to have it appear other than an image for it but that sounds like an awful solution so I'll think about it
image

TGCrystal:
> PDF seems even worse in nightly so I would call this an improvement – also this is such a small use-case (like who the hell needs a PDF form of their circuit on their phones) so I'm gonna say it's out of scope

Fair enough


> Definitely allowing them to press "F" is a good plan - do you think there should be a physical button too?

Absolutely, particularly for mobile.


> You're talking about this pattern? Yeah I couldn't find a trivial way to have it appear other than an image for it but that sounds like an awful solution so I'll think about it

Yeah I figured it probably wasn't super easy to make and that's why it wasn't there already. Definitely fair to call it outside the scope of this pr if it isn't trivial.

TGCrystal (approved):

Issue #825, opened 2021-10-29
Using the snip tool on colored wires makes the resulting wire white


OwenMcGee:
**Describe the bug**
When using the snip tool on wires that are colored, the resultant wire (where there were previously two (colored) segments) will be white.

**To Reproduce**
1. Add a switch and an LED to the canvas
2. Connect the two with a wire, and add a port in the middle of the wire
3. Color both segments of the wire some color other than white
4. Select the port in the middle of the wire, and press x to snip
5. Observe the resultant white wire

**Expected behavior**
If the two wire segments were the same color, I would expect the resultant wire to also be that color. If the two wires were _not_ the same color, I'm unsure what the best behavior would be. Perhaps take the color of one of the two wires at random, or maybe take the average color, or something else.

**Screenshots**
The current behavior when snipping colored wires:
![Animation2](https://user-images.githubusercontent.com/87297165/139502653-1cdfd317-2de4-4276-a547-4bf43f5bab76.gif)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome

PR #919, opened 2021-11-26
Fixed snipped wires coloring 

vhada7:
When wires with a port are colored something other than white, the color will remain the same even after the port is snipped. If the two split wires are different colors then the resulting wire after snipping will have a color that is a blend of the previous two
colors (using blend function). 

Fixes issue #825


OwenMcGee:
This seems good! I like the color blending. One thing I'll note is that if you select multiple ports on the same wire and snip, it snips them one at a time, which means that the color doesn't get blended evenly. Not a result of your code though, just the existing functionality.

seal of approval

LeonMontealegre (approved):

LeonMontealegre (approved):

Issue #934, opened 2021-12-05
Circuit to Expression


TGCrystal:
**Is your feature request related to a problem? Please describe.**
Recently, an expression to circuit feature was added. Currently, there is no way to do the opposite of this. 

**Describe the solution you'd like**
Reusing slightly modified data structures from the expression to circuit feature, it shouldn't be too hard to go the other way.

**Additional context**
Design doc with more details coming soon.

PR #945, opened 2022-01-01
Expression parser improvements

### LINKED ISSUE #897: When using the expression to circuit feature, add option to create with clocks set up as a truth table
Clocks can be used to simulate a truth table for a circuit by having clocks connected to each input, with each subsequent clock having twice the delay of the last. This can be seen by connecting all inputs (and likely, the output as well) to an oscilloscope. An example of this is shown below (in retrospect this is a somewhat poor example as a boolean expression can only result in a single output, but just ignore that):
image
Or, to steal Professor Kuzmin's image to see how this oscilloscope is a truth table:
image

I believe it would be a useful feature to, when selecting clocks as the input component when using the expression to circuit feature, allow the user to have the clocks be created in this manner (each having double the delay of the last, and potentially connecting all inputs and outputs to an oscilloscope (or perhaps one for inputs and one for outputs, to avoid the messy connections as in my example).

I am actually hard pressed to imagine a circumstance when I would want clocks as inputs but all at the same delay and in sync, so I may even suggest making this the default behavior with clock inputs, but that's up for debate. An additional checkbox for "Create as truth table with oscilloscopes" or something is also fine.

Some issues I can predict if this was to be added:
1. There is a minimum and maximum to the amount of delay a clock can have, meaning that at most 8 clocks can be placed until it is impossible to double the delay again. I'm not sure what the best solution would be for expressions with over 8 inputs. A clock with above maximum delay could be simulated with some extra components in that section, but that might be doing too much.
2. Oscilloscopes only allow 8 inputs, so circuits with more would run into an issue trying to fit more. Although two oscilloscopes right on top of one another looks practically identical to simply having a taller oscilloscope, so that could be a way around it
3. If the oscilloscope connects both inputs and outputs, the wiring has the potential to get messy quickly, unless extra thought is put into it. Even if separate oscilloscopes are used, ideally they would be close to one another so they could actually be used to see what inputs result in what outputs.

Additional notes:
- We also need to make sure that the oscilloscope's delay isn't larger than that of the quickest clock
- The oscilloscopes being present isn't strictly necessary, and even without them this would still be more useful than the current clock inputs in my opinion

### LINKED ISSUE #909: Add option to include labels when creating a circuit from a boolean expression
When using the expression to circuit feature, all inputs are already labelled by necessity in the expression, and it could be helpful to the user to display a label next to the switches (or buttons or clocks) when they appear on the canvas. A checkbox asking "Add labels next to inputs?" on the righthand side would be sufficient.

If the "Generate into IC" option is checked, ~this checkbox should probably be forced to be off, since the inputs are already visibly labelled~ these labels should be generated for the _internal_ circuit, but not placed on the canvas directly. That is, when choosing "View IC" on the resulting IC, the circuit that appears should have the labels.

### LINKED ISSUE #927: More helpful error messages in the expression to circuit feature
If an invalid expression is given, we could quickly parse through and check if the expression contains any operands from a notation set not currently selected. If the expression _does_ contain any of those operands, we could add an additional "Make sure you're using the right operands!", or "Remember that you can only use the operands from a single notation set at a time" message to give the user more direction. Perhaps the message should specify which of these incorrect operands it is finding.

(an example in which this would have helped someone that I've seen, was a user attempting to enter the expression "(A and B) OR (C and D)", while having the Literal 2 notation set selected (meaning the "OR" was invalid)).

TGCrystal:
closes #927 
closes #909 
closes #897 

Also includes some under the hood improvements to make more parts of the pipeline action based. This helps future proof with the proposed collaborative editing feature which requires pretty much everything to be an action. It also will make converting some of the currently frontend only code to an action which eases testability. Currently, I am waiting on an action based circuit arrangement algorithm to fully go through with that.

There are also some changes to the tree structure of the underlying algorithm. That now directly handles the existence of NAND, NOR, and XNOR gates. Also, it can handle OR, AND, and XOR gates of different sizes. This helps to pave the way for #934 

Some notes concering #897: To try to balance what looked best on the oscilloscope, I did slightly modify the intervals from what was mentioned in the issue. Now, there are 5 different times (from 500 to 8000) that the clocks can be set to instead of 7. Also, I realized our small oversight that you can't actually connect multiple things to one oscilloscope import, so inputs past the first 5 will not be connected to the oscilloscope.

Lastly, I know the oscilloscope placement doesn't look great. I'm not fixing it.


src/app/digital/utils/ExpressionParser/GenerateInputTree.ts
```diff
@@ -76,27 +78,60 @@ const DefaultPrecedences: TokenType[] = ["|", "^", "&", "!", "("];
...
+    }
+    else if (currentOp === "|" || currentOp === "^" || currentOp === "&") {
+        const lTree = leftRet.tree;
+        const rTree = rightRet.tree;
+        // 7 + rTree = 8, that is why 7 is used
+        let childrenArray: InputTree[] = (lTree.kind === "binop"
+                             && lTree.type === currentOp
+                             && !lTree.isNot
+                             && !leftRet.final
+                             && lTree.children.length <= 7) ? lTree.children : [lTree];
```
LeonMontealegre, line 117: Could also just do `< 8` for clarity right?

LeonMontealegre, line 117: This is a lot of conditions, what exactly are you checking for? Might be better in a utility function like `isTreeBLAH()` especially since a lot of the conditions are repeated for `rTree` below

src/app/digital/utils/ExpressionParser/GenerateInputTree.ts
```diff
@@ -76,27 +78,60 @@ const DefaultPrecedences: TokenType[] = ["|", "^", "&", "!", "("];
...
+        if (rTree.kind === "binop" && rTree.type === currentOp && !rTree.isNot && !rightRet.final) {
+            const rChildren: InputTree[] = rTree.children;
+            while (rChildren.length > 0 && childrenArray.length < 7)
+                childrenArray.push(rChildren.shift());
```
LeonMontealegre, line 121: are you just adding all the children in `rTree.children` to `childrenArray`? Can just do `childrenArray = [...childrenArray, ...rTree.children];`

LeonMontealegre: Or I guess you're only adding the first 7? Can just do same but with `rTree.children.slice(0, 7)`

src/app/digital/utils/ExpressionParser/GenerateInputTree.ts
```diff
@@ -124,15 +159,25 @@ export function GenerateInputTree(tokens: Token[], ops: Record<TokenType, string
...
+            let errorMessage = `No valid operator between "${prev.name}" and "${next.name}"`;
+            for (const format of Formats) {
+                Object.entries(format.ops).forEach(([tokenType, op]) => {
+                    if (op === prev.name)
+                        errorMessage += `\nDid you mean to use "${ops[tokenType as TokenType]}" instead of "${prev.name}"?`
+                    if (op === next.name)
+                        errorMessage += `\nDid you mean to use "${ops[tokenType as TokenType]}" instead of "${next.name}"?`
+                });
+            }
+            throw new Error(errorMessage);
```
LeonMontealegre, line 179: This function is starting to get quite long, hopefully the previous suggestions will lead to a smaller size, but also here you could probably make a `BuildErrorMessage` function or something and keep the `throw new Error` as a 1-liner

src/site/pages/digital/src/containers/ExprToCircuitPopup/generate.ts
```diff
@@ -1,73 +1,140 @@
...
-    if (format !== "custom")
-        ops = Formats.find(form => form.icon === format);
+                         options: ExprToCirGeneratorOptions) {
+    const input = options.input ?? "Switch";
+    const output = options.output ?? "LED";
+    const isIC = (output !== "Oscilloscope") ? (options.isIC ?? false) : false;
+    const connectClocksToOscope = options.connectClocksToOscope ?? false;
+    const label = options.label ?? false;
+    const format = options.format ?? "|";
+    const ops = (format === "custom") ? (options.ops ?? Formats[0]) : Formats.find(form => form.icon === format);
```
LeonMontealegre, line 54: you can define a `defaultOptions` constant above and then just have `const options = {...defaultOptions, ...userOptions}` and then also make `type ExprToCirGeneratorOptions` not have optional parameters and just make the parameter `userOptions: Partial<ExprToCirGeneratorOptions>`

Also if you want all the options to just be accessible by name without like `options.input` you can do
```ts
const {input, output, isIC, ...} = {...defaultOptions, ...userOptions};
```
but up to you, I think I would just go for using `options.whatever` tho

src/site/pages/digital/src/containers/ExprToCircuitPopup/generate.ts
```diff
@@ -1,73 +1,140 @@
...
+
+    const circuitComponents = circuit.getComponents();
+
+    // Add labels next to inputs
+    // TODO: This will have to be redone when there is a better organization algorithm
+    if (label) {
+        for (const [name, component] of inputMap) {
+            const newLabel = Create<Label>("Label");
+            // 100 is an arbitrary value here
+            const pos = component.getPos().sub(100, 0);
```
LeonMontealegre, line 96: `sub` width of `component` instead?

src/site/pages/digital/src/containers/ExprToCircuitPopup/generate.ts
```diff
@@ -1,73 +1,140 @@
...
+            action.add(new TranslateAction([newLabel], [newLabel.getPos()], [pos]).execute());
+            circuitComponents.push(newLabel);
+        }
+    }
+
+    // Set clock frequencies, also connect to oscilloscope if that option is set
+    if (input === "Clock") {
+        let inIndex = 0;
+        // Set clock frequencies
+        for (let clock of inputMap.values() as IterableIterator<Clock>) {
```
LeonMontealegre, line 108: this cast is necessary? I've never heard of `IterableIterator` before

src/site/pages/digital/src/containers/ExprToCircuitPopup/generate.ts
```diff
@@ -1,73 +1,140 @@
...
+            const pos = component.getPos().sub(100, 0);
+            action.add(new PlaceAction(info.designer, newLabel).execute());
+            action.add(new SetNameAction(newLabel, name).execute());
+            action.add(new TranslateAction([newLabel], [newLabel.getPos()], [pos]).execute());
+            circuitComponents.push(newLabel);
+        }
+    }
+
+    // Set clock frequencies, also connect to oscilloscope if that option is set
+    if (input === "Clock") {
```
LeonMontealegre, line 105: I think all these post-process if-statements should be moved into their own function

src/site/pages/digital/src/containers/ExprToCircuitPopup/generate.ts
LeonMontealegre, line 108: also use `const` not `let` since `clock` isn't changing

src/site/pages/digital/src/containers/ExprToCircuitPopup/generate.ts
```diff
@@ -1,73 +1,140 @@
...
+        // Set clock frequencies
+        for (let clock of inputMap.values() as IterableIterator<Clock>) {
+            action.add(new FrequencyChangeAction(clock, 500 * (2 ** inIndex)).execute());
+            inIndex = Math.min(inIndex+1, 4);
+        }
+        // Connect clocks to oscilloscope
+        if (connectClocksToOscope) {
+            inIndex = 0;
+            action.add(new InputPortChangeAction(o, 1, Math.min(inputMap.size + 1, 6)).execute());
+            for (let clock of inputMap.values() as IterableIterator<Clock>) {
```
LeonMontealegre, line 116: use `const`

src/site/pages/digital/src/containers/ExprToCircuitPopup/index.tsx
```diff
@@ -75,27 +82,44 @@ export const ExprToCircuitPopup = (({ mainInfo }: Props) => {
...
                 <button className="exprtocircuit__popup__generate" type="button" disabled={expression===""} onClick={() => {
                     try {
-                        Generate(mainInfo, expression, isIC, input, format, customOps);
+                        const options: ExprToCirGeneratorOptions = {
+                            input: input,
+                            output: output,
+                            isIC: isIC,
+                            connectClocksToOscope: clocksToOscope,
+                            label: label,
+                            format: format,
```
LeonMontealegre, line 119: most of these you can shorthand as just `label, format, input, output, etc`

LeonMontealegre (changes requested):

src/app/digital/utils/ExpressionParser/GenerateInputTree.ts
```diff
@@ -1,15 +1,68 @@
...
+ * @returns the properly nested tree structure
+ */
+function generateNestedTrees(children: InputTree[], currentOp: InputTreeBinOpType): InputTree[] {
+    if (children.length <= 8)
+        return children;
+    const next = children.slice(7);
+    const newTree: InputTree = {
+        kind: "binop", type: currentOp, isNot: false,
+        children: generateNestedTrees(next, currentOp) as BinOpChildren,
+    }
```
LeonMontealegre, line 41: nit: `;`

src/app/digital/utils/ExpressionParser/GenerateInputTree.ts
```diff
@@ -76,27 +129,48 @@ const DefaultPrecedences: TokenType[] = ["|", "^", "&", "!", "("];
...
+            tree = {kind: "unop", type: "!", child: rightRet.tree};
+        }
+    }
+    else if (currentOp === "|" || currentOp === "^" || currentOp === "&") {
+        const lTree = leftRet.tree;
+        const rTree = rightRet.tree;
+        let childrenArray: InputTree[] = (isTreeExtendable(lTree, currentOp, leftRet.final)) ? lTree.children : [lTree];
+        if (isTreeExtendable(rTree, currentOp, rightRet.final))
+            childrenArray = [...childrenArray, ...rTree.children];
+            //a|b|c|d|e|f|g|h|i|j|k|l|m|n|o|p|q|r|s|t|u|v|w|x|y|z
```
LeonMontealegre, line 166: ?

src/app/digital/utils/ExpressionParser/GenerateInputTree.ts
```diff
@@ -76,27 +129,48 @@ const DefaultPrecedences: TokenType[] = ["|", "^", "&", "!", "("];
...
+            tree.isNot = true;
+        }
+        else {
+            tree = {kind: "unop", type: "!", child: rightRet.tree};
+        }
+    }
+    else if (currentOp === "|" || currentOp === "^" || currentOp === "&") {
+        const lTree = leftRet.tree;
+        const rTree = rightRet.tree;
+        let childrenArray: InputTree[] = (isTreeExtendable(lTree, currentOp, leftRet.final)) ? lTree.children : [lTree];
```
LeonMontealegre, line 163: Do you need the explicit type here?

src/app/digital/utils/ExpressionParser/GenerateInputTree.ts
```diff
@@ -76,27 +129,48 @@ const DefaultPrecedences: TokenType[] = ["|", "^", "&", "!", "("];
...
+            tree = rTree;
+            tree.isNot = true;
+        }
+        else {
+            tree = {kind: "unop", type: "!", child: rightRet.tree};
+        }
+    }
+    else if (currentOp === "|" || currentOp === "^" || currentOp === "&") {
+        const lTree = leftRet.tree;
+        const rTree = rightRet.tree;
```
LeonMontealegre, line 162: you can combine these to like `const lTree = leftRet.tree, rTree = rightRet.tree`

src/site/pages/digital/src/containers/ExprToCircuitPopup/generate.ts
```diff
@@ -1,73 +1,154 @@
...
+        action.add(new SetNameAction(ic, expression).execute());
         action.add(new CreateICDataAction(data, info.designer).execute());
+        action.add(CreateDeleteGroupAction(info.designer, circuitComponents).execute());
         action.add(new PlaceAction(info.designer, ic).execute());
+        action.add(new TranslateAction([ic], [ic.getPos()], [info.camera.getPos()]).execute());
         action.add(new SelectAction(info.selections, ic).execute());
     } else { // If placing directly
-        action.add(CreateGroupSelectAction(info.selections, negated.getComponents()).execute());
+        action.add(CreateGroupSelectAction(info.selections, circuitComponents).execute());
     }
```
LeonMontealegre, line 152: I think also add this one to its own function

And also I think you can just pass in `options` to each function and they can check themselves for `if (!options.isIC) return false;` for example

src/site/pages/digital/src/containers/ExprToCircuitPopup/index.tsx
```diff
@@ -75,27 +81,41 @@ export const ExprToCircuitPopup = (({ mainInfo }: Props) => {
...
                 <button className="exprtocircuit__popup__generate" type="button" disabled={expression===""} onClick={() => {
                     try {
-                        Generate(mainInfo, expression, isIC, input, format, customOps);
+                        const options: ExprToCirGeneratorOptions = {
+                            input, output, isIC,
+                            connectClocksToOscope: clocksToOscope,
+                            label, format,
+                            ops: customOps,
+                        }
+                        Generate(mainInfo, expression, options);
```
LeonMontealegre, line 118: nit: have it like
```ts
                        Generate(mainInfo, expression, {
                            input, output, isIC,
                            connectClocksToOscope: clocksToOscope,
                            label, format,
                            ops: customOps,
                        });
```

LeonMontealegre: (unless you're using `options` somewhere else down there that i can't see)

LeonMontealegre (changes requested):

LeonMontealegre (approved):

PR #954, opened 2022-01-12
Fixed issue with initial IC propagation

### LINKED ISSUE #468: ICs don't propagate signals immediately when connected

LeonMontealegre:
Fixes #468

### DIFF
--- a/src/app/tests/digital/models/ioobjects/other/IC.test.ts
+++ b/src/app/tests/digital/models/ioobjects/other/IC.test.ts
@@ -1,6 +1,7 @@
 import "jest";
 
 import {DigitalCircuitDesigner} from "digital/models/DigitalCircuitDesigner";
+import {ConstantHigh}    from "digital/models/ioobjects/inputs/ConstantHigh";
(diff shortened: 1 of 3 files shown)


src/app/tests/digital/models/ioobjects/other/IC.test.ts
```diff
@@ -10,6 +10,7 @@ import {IC}     from "digital/models/ioobjects/other/IC";
 import {ICData} from "digital/models/ioobjects/other/ICData";
 
 import {GetHelpers} from "test/helpers/Helpers";
+import {ConstantHigh} from "digital/models/ioobjects";
```
TGCrystal, line 13: Is this import order correct?

TGCrystal (changes requested):

TGCrystal (approved):

PR #986, opened 2022-02-10
Copy from component popup

### LINKED ISSUE #874: Copy pasting text from a component's selection popup pastes the entire component
**Describe the bug**
When copying a field from a selection popup (such as the position, delay, or component name), if the user attempts to paste the information into another input field the site will act as if the user was just copy pasting the component, not the text.

**To Reproduce**
1. Add two labels to the canvas
2. Give one label a new name ("Input", for example)
3. Highlight the label's new name in the selection popup, and copy it
4. Attempt to paste the name into the other label's name field
5. observe error

**Expected behavior**
I would expect that when text is highlighted by the user, nothing but that text is copied to the clipboard.

**Screenshots**
A gif showing an example of this happening (the same example as described above):
animation

 - OS: Windows 10
 - Browser: Chrome

**Additional context**
When doing the original copying from somewhere else (some other tab), the pasting works fine. It appears that this bug is entirely an issue with copying, not with pasting.

riedee:
Resolves #874

If any text is selected within a component's popup, it will copy the text instead of the entire component as it did previously.
If the cursor is in the text field but nothing is selected, it will copy the component. I thought this made the most sense, but that can be changed easily if not.

### DIFF
--- a/src/app/core/tools/handlers/CopyHandler.ts
+++ b/src/app/core/tools/handlers/CopyHandler.ts
@@ -14,7 +14,14 @@ import {EventHandler} from "../EventHandler";
 export const CopyHandler: EventHandler = ({
     conditions: (event: Event, {selections}: CircuitInfo) =>
         ((event.type === "copy" || event.type === "cut") &&
-         selections.amount() > 0),
+         selections.amount() > 0) &&
+        // If a label element is not selected, 
+        // or if an editable component is selected but it is a "Caret" (cursor) instead of highlighted text,
+        // then copy the component
+        // Otherwise there is text selected, so do default copying
+        // Necessary to fix #874
+        (document.getSelection()?.anchorNode?.nodeName !== "LABEL" || 
+         document.getSelection()?.type === "Caret"),
 
     getResponse: ({selections, designer, history}: CircuitInfo, {type, ev}: CopyPasteEvent) => {
         const objs = selections.get().filter(o => o instanceof IOObject) as IOObject[];
@@ -27,16 +34,8 @@ export const CopyHandler: EventHandler = ({
         // We don't copy the data from the json since it will cause 
         // some weird error, which will cause the issue #746
         ev.clipboardData.setData("text/plain", str);
-
-        // If an editable component (LABEL) is not selected, 
-        // or if an editable component is selected but it is a "Caret" (cursor) instead of highlighted text,
-        // then copy the component
-        // Otherwise there is text selected, so do default copying
-        // Necessary to fix #874
-        if (document.getSelection()?.anchorNode?.nodeName !== "LABEL" || document.getSelection()?.type === "Caret") {
-            ev.preventDefault(); // Necessary to copy correctly
-        }
-
+        ev.preventDefault(); // Necessary to copy correctly
+        
         if (type === "cut") {
             // Delete selections
             history.add(new GroupAction([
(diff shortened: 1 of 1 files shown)


src/app/core/tools/handlers/CopyHandler.ts
```diff
@@ -27,7 +27,12 @@ export const CopyHandler: EventHandler = ({
         // We don't copy the data from the json since it will cause 
         // some weird error, which will cause the issue #746
         ev.clipboardData.setData("text/plain", str);
-        ev.preventDefault(); // Necessary to copy correctly
+
+        // If no text is selected, copy the component
+        // Otherwise, do default copying
```
LeonMontealegre, line 32: Can you expand this comment to explain what exactly this if-statement is doing?
I'm not sure what the `'LABEL'` thing represents, or why the `"CARET"` matters?

Also, you should say (in the comment) that this statement is necessary because it "fixes #874"

src/app/core/tools/handlers/CopyHandler.ts
```diff
@@ -27,7 +27,12 @@ export const CopyHandler: EventHandler = ({
...
+        if (document.getSelection()?.anchorNode?.nodeName != 'LABEL' || document.getSelection()?.type == "Caret") {
```
LeonMontealegre, line 33: nit: use `!==`
nit: use `"` instead of `'` for strings
nit: use `===`

LeonMontealegre (changes requested):

src/app/core/tools/handlers/CopyHandler.ts
riedee: Yeah that definitely wasn't clear, hopefully the comments I added make it more understandable.

src/app/core/tools/handlers/CopyHandler.ts
```diff
@@ -27,7 +27,15 @@ export const CopyHandler: EventHandler = ({
         // We don't copy the data from the json since it will cause 
         // some weird error, which will cause the issue #746
         ev.clipboardData.setData("text/plain", str);
-        ev.preventDefault(); // Necessary to copy correctly
+
+        // If an editable component (LABEL) is not selected, 
+        // or if an editable component is selected but it is a "Caret" (cursor) instead of highlighted text,
```
LeonMontealegre, line 32: Definitely clearer, but where exactly did you get the string `"LABEL"`, why is it in all caps? I've never seen it before so I'm more just wondering.

riedee: Basically I noticed that all the input elements in the popup modules had that anchor node name. I believe it's because in Module.tsx there's this section for the popup module:
```
export const PopupModule = (({label, modules}: PopupModuleProps) => {
    return function PopupModuleFunc (props: UseModuleProps) {
        const ms = modules.map(m => m(props));
        if (ms.every(m => m === null))
            return null;
        return <div key={`selection-popup-${label}-module`}>
            {label}
            <label unselectable="on">
                {ms.map((m, i) => cloneElement(m!, {key: `selection-popup-${label}-module-${i}`}))}
            </label>
        </div>;
    }
});
```
Each of the inputs seems to be surrounded by a label element, which is why when my cursor is in an input and I call the anchor node name, it returns label. I'm not sure why it's in all caps, but it seems to be the same for all other types of node names as well (I'm assuming it's just an HTML thing?).

LeonMontealegre (approved):

PR #987, opened 2022-02-10
Add ability to double click to select all connected components

### LINKED ISSUE #800: Double click on a component to select all connected components
Similar to how double clicking on a wire will select all of the wire's components and nodes splitting it, it could be useful to be able to double click a component and select all other components connected to it.

This would essentially select all the components necessary to create an IC containing the component that was double clicked, which could be easier than selecting those components individually (particularly if the canvas contains many other, non-connected components in the way, making bounding box highlighting more tedious. This also would seem to be the easiest method for users on a trackpad, without a mouse).

mercuryhg31:
Closes #800

### DIFF
--- a/src/app/core/tools/handlers/SelectPathHandler.ts
+++ b/src/app/core/tools/handlers/SelectPathHandler.ts
@@ -6,9 +6,10 @@ import {GetComponentPath, GetWirePath} from "core/utils/ComponentUtils";
 
 import {CreateGroupSelectAction} from "core/actions/selection/SelectAction";
 
+import {Component, Wire} from "core/models";
+
 import {EventHandler} from "../EventHandler";
 
-import {Component, Wire} from "core/models";
 
 export const SelectPathHandler: EventHandler = ({
     conditions: (event: Event, {input, camera, designer}: CircuitInfo) =>
--- a/src/app/core/tools/handlers/SelectPathHandler.ts
+++ b/src/app/core/tools/handlers/SelectPathHandler.ts
@@ -8,43 +8,30 @@ import {CreateGroupSelectAction} from "core/actions/selection/SelectAction";
 
 import {EventHandler} from "../EventHandler";
 
+import {Component} from "core/models";
+import {Wire} from "core/models";
 
 export const SelectPathHandler: EventHandler = ({
     conditions: (event: Event, {input, camera, designer}: CircuitInfo) =>
         (event.type === "dblclick" &&
          event.button === LEFT_MOUSE_BUTTON &&
-         ( // is there a wire or component within select bounds?
-          designer
-            .getWires()
-            .find(o =>
-                o.isWithinSelectBounds(camera.getWorldPos(input.getMousePos())))
-            !== undefined ||
-          designer
-            .getObjects()
-            .find(o =>
-                o.isWithinSelectBounds(camera.getWorldPos(input.getMousePos())))
-            !== undefined
-         )
+         designer // is there a wire or component within select bounds?
+           .getAll()
+           .find(o =>
+               o.isWithinSelectBounds(camera.getWorldPos(input.getMousePos())))
+           !== undefined
         ),
 
     getResponse: ({input, camera, history, designer, selections}: CircuitInfo) => {
         const worldMousePos = camera.getWorldPos(input.getMousePos());
 
-        const wire = designer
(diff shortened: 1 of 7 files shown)


TGCrystal (approved):

src/app/core/utils/ComponentUtils.ts
```diff
@@ -136,6 +136,38 @@ export function GetPath(w: Wire | Node): Array<Wire | Node> {
...
+    const queue = new Array<Component>(c);
+    const visited = new Set<Component>();
+
+    while(queue.length > 0) {
+        const q = queue.shift()!;
+
+        visited.add(q);
+        path.push(q);
+        for (let p of q.getPorts()) {
+            for (let w of p.getWires()) {
```
LeonMontealegre, line 159: nit: use `const`

src/app/core/tools/handlers/SelectPathHandler.ts
```diff
@@ -27,8 +34,17 @@ export const SelectPathHandler: EventHandler = ({
...
+            .find(o => o.isWithinSelectBounds(worldMousePos));
 
-        history.add(CreateGroupSelectAction(selections, path).execute());
+        if (wire != null) {
+            const wirePath = GetWirePath(wire!);
+            history.add(CreateGroupSelectAction(selections, wirePath).execute());
+        } else {
+            const objPath = GetComponentPath(component!);
+            history.add(CreateGroupSelectAction(selections, objPath).execute())
+        }
```
LeonMontealegre, line 48: Can condense this as
```ts
const path = (wire) ? (GetWirePath(wire)) : (GetComponentPath(component!));
history.add(CreateGroupSelectAction(selections, path).execute());
```

src/app/core/tools/handlers/SelectPathHandler.ts
```diff
@@ -27,8 +34,17 @@ export const SelectPathHandler: EventHandler = ({
             .getWires()
             .reverse()
             .find(o => o.isWithinSelectBounds(worldMousePos));
-        const path = GetPath(wire!);
+        const component = designer
+            .getObjects()
+            .reverse()
+            .find(o => o.isWithinSelectBounds(worldMousePos));
```
LeonMontealegre, line 40: Combine these using `getAll()`, which, when reversed, will have the wires in front and the objects at the back which is the desired behavior. (also make sure to cast `as (Component | Wire)`)

Then for the if-statement below, use `(obj instanceof Wire)`

LeonMontealegre (changes requested):

src/app/core/tools/handlers/SelectPathHandler.ts
```diff
@@ -2,33 +2,36 @@ import {LEFT_MOUSE_BUTTON} from "core/utils/Constants";
...
+               o.isWithinSelectBounds(camera.getWorldPos(input.getMousePos())))
+           !== undefined
+        ),
 
     getResponse: ({input, camera, history, designer, selections}: CircuitInfo) => {
         const worldMousePos = camera.getWorldPos(input.getMousePos());
 
-        const wire = designer
-            .getWires()
+        const obj: Component | Wire = designer
```
LeonMontealegre, line 28: nit: Shouldn't need a double cast, the cast below should be enough

src/app/core/tools/handlers/SelectPathHandler.ts
```diff
@@ -2,33 +2,36 @@ import {LEFT_MOUSE_BUTTON} from "core/utils/Constants";
...
 import {CircuitInfo} from "core/utils/CircuitInfo";
-import {GetPath} from "core/utils/ComponentUtils";
+import {GetComponentPath, GetWirePath} from "core/utils/ComponentUtils";
 
 import {CreateGroupSelectAction} from "core/actions/selection/SelectAction";
 
 import {EventHandler} from "../EventHandler";
 
+import {Component} from "core/models";
+import {Wire} from "core/models";
```
LeonMontealegre, line 12: nit: Combine these to `import {Component, Wire} from "core/models";` and more the import to be above the ".." import

LeonMontealegre (changes requested):
Just a few nits!

TGCrystal (approved):

LeonMontealegre (approved):

PR #988, opened 2022-02-10
Allow optional disabling of snapping when using the Translate Action

### LINKED ISSUE #910: Snapping components to connected ports' locations overriding manual input
**Describe the bug**
When a user attempts to manually input a component's position via the horizontal and vertical position fields, if the new location is close enough to snap to a nearby connected component's location, it will do so and ignore the manual input.

**To Reproduce**
1. Add two components connected to one another to the canvas (i.e. a switch and an LED)
2. Arrange the components such that they are snapped together, and take note of the position (in the axis they are snapped in)
3. Move one component away, and then manually set its position (in the relevant axis) to what it used to be, except greater by 0.1
4. Select the component again and observe that it was moved to its original position, despite being different than the user input

**Expected behavior**
I would expect the snapping behavior to be overridden whenever the user manually enters positions into the selection popup, or whenever the user holds shift to snap to gridlines.

 - OS: Windows 10
 - Browser: Chrome

**Additional context**
This was mildly fixed with something of a workaround to solve issue #615, but I think it's still pressing enough to deserve a more thorough fix. In particular, this issue becomes quite annoying when trying to slightly move a component inside an existing circuit, as the only way to do so without first deleting all its connections is to move every port/component it is snapped with away temporarily, move the component in question, and then move all the other components back again, which is quite tedious.

Cyphecx:
Fixes #910
I added an optional "snap" flag to the constructor of the Translate Action. All of the snapping calls are ignored when this is set to false. 
This addresses the issue in #910 by setting the translate constructors snap flag to false in the postion module of the selection pop up. This prevents snapping when making minute adjustments to position in the selection pop up. 

Additionaly fixed some minor linting issues.

### DIFF
--- a/src/app/core/actions/transform/TranslateAction.ts
+++ b/src/app/core/actions/transform/TranslateAction.ts
@@ -28,6 +28,8 @@ export class TranslateAction implements Action {
      */
     protected targetPositions: Vector[];
 
+    protected snap: boolean;
+
     /**
      * Creates a translation of component(s) from one position to another.
      * Each component in objs list has corresponding initial position and target position in those
@@ -37,11 +39,13 @@ export class TranslateAction implements Action {
      * @param initialPositions Initializes the array with the selected components' starting positions
      * @param targetPositions Initializes the array with the selected components' final positions
      */
-    public constructor(objs: Component[], initialPositions: Vector[], targetPositions: Vector[]) {
+    public constructor(objs: Component[], initialPositions: Vector[], targetPositions: Vector[], snap?: boolean) {
         this.objs = objs;
 
         this.initialPositions = initialPositions;
         this.targetPositions = targetPositions;
+        if(typeof snap === "undefined"){ this.snap = true; }
+        else{ this.snap = snap;}
     }
 
     /**
@@ -53,7 +57,9 @@ export class TranslateAction implements Action {
         this.objs.forEach((o, i) => o.setPos(this.targetPositions[i]));
 
         // Always snap afterwards to avoid issue #417
-        this.objs.forEach(o => SnapPos(o));
+        if(this.snap){
+            this.objs.forEach(o => SnapPos(o));
+        }
 
         return this;
     }
@@ -67,7 +73,9 @@ export class TranslateAction implements Action {
         this.objs.forEach((o, i) => o.setPos(this.initialPositions[i]));
 
         // Always snap afterwards to avoid issue #417
-        this.objs.forEach(o => SnapPos(o));
+        if(this.snap){
+            this.objs.forEach(o => SnapPos(o));
+        }
 
         return this;
     }
--- a/src/app/core/actions/transform/TranslateAction.ts
+++ b/src/app/core/actions/transform/TranslateAction.ts
@@ -9,8 +9,7 @@ import {SnapPos} from "./SnapUtils";
 
 /**
  * Translate can be applied to single components or groups of components,
- * used for moving componets from one position to another.* used for moving
- * componets from one position to another.
+ * used for moving componets from one position to another.
  */
 export class TranslateAction implements Action {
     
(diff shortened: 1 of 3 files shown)


src/app/core/actions/transform/TranslateAction.ts
```diff
@@ -37,11 +39,13 @@ export class TranslateAction implements Action {
...
      * @param targetPositions Initializes the array with the selected components' final positions
      */
-    public constructor(objs: Component[], initialPositions: Vector[], targetPositions: Vector[]) {
+    public constructor(objs: Component[], initialPositions: Vector[], targetPositions: Vector[], snap?: boolean) {
         this.objs = objs;
 
         this.initialPositions = initialPositions;
         this.targetPositions = targetPositions;
+        if(typeof snap === "undefined"){ this.snap = true; }
+        else{ this.snap = snap;}
```
TGCrystal, line 48: This syntax can be cleaned up by changing `snap?: boolean` to `snap = true`. Then just `this.snap = snap;` is all that is necessary below.

src/app/core/actions/transform/TranslateAction.ts
```diff
@@ -53,7 +57,9 @@ export class TranslateAction implements Action {
         this.objs.forEach((o, i) => o.setPos(this.targetPositions[i]));
 
         // Always snap afterwards to avoid issue #417
-        this.objs.forEach(o => SnapPos(o));
+        if(this.snap){
+            this.objs.forEach(o => SnapPos(o));
+        }
```
TGCrystal, line 62: `{}` are not needed for single line if statements like this one

src/app/core/actions/transform/TranslateAction.ts
```diff
@@ -67,7 +73,9 @@ export class TranslateAction implements Action {
         this.objs.forEach((o, i) => o.setPos(this.initialPositions[i]));
 
         // Always snap afterwards to avoid issue #417
-        this.objs.forEach(o => SnapPos(o));
+        if(this.snap){
+            this.objs.forEach(o => SnapPos(o));
+        }
```
TGCrystal, line 78: `{}` are not needed for single line if statements like this one

TGCrystal (changes requested):
This fixes the case when editing the value in the text box, but not when holding shift to snap to the grid lines. That snapping behavior is still being overridden by snapping to other port.

TGCrystal (approved):

src/app/core/actions/transform/TranslateAction.ts
```diff
@@ -53,7 +56,8 @@ export class TranslateAction implements Action {
         this.objs.forEach((o, i) => o.setPos(this.targetPositions[i]));
 
         // Always snap afterwards to avoid issue #417
-        this.objs.forEach(o => SnapPos(o));
+        if(this.snap)
```
LeonMontealegre, line 59: nit: space after `if`

src/app/core/actions/transform/TranslateAction.ts
```diff
@@ -67,7 +71,8 @@ export class TranslateAction implements Action {
         this.objs.forEach((o, i) => o.setPos(this.initialPositions[i]));
 
         // Always snap afterwards to avoid issue #417
-        this.objs.forEach(o => SnapPos(o));
+        if(this.snap)
```
LeonMontealegre, line 74: nit: space after `if`

src/app/core/tools/TranslateTool.ts
```diff
@@ -104,8 +104,9 @@ export const TranslateTool: Tool = (() => {
                         curPositions.map(p => Snap(p)):
                         curPositions;
 
+                    const snapping = input.isShiftKeyDown() ? false : true;
```
LeonMontealegre, line 107: This sounds unintuitive because generally holding shift would be snapping, let's call this `snapToConnections`

LeonMontealegre (changes requested):
just some nits and formatting stuff!

LeonMontealegre (approved):

TGCrystal (approved):

PR #996, opened 2022-02-11
Add history box button to header

### LINKED ISSUE #979: History Box: Add another way to open the history box
The history box currently can be opened only through a button in the itemnav. The history box also is not closed when the circuit is locked (which is the correct, intended behavior).
Those two facts cause a weird incongruity in that it is not possible to open the history box while the circuit is locked, even though it is allowed.
There should be another button somewhere else (probably either the header or the sidenav) to open the history box to correct this.

EricW123:
Fixes #979.

Added a new history button at top left corner, next to locker button.


src/site/shared/containers/Header/Left/Locker.tsx
```diff
@@ -0,0 +1,24 @@
+import {useSharedDispatch, useSharedSelector} from "shared/utils/hooks/useShared";
+import {ToggleCircuitLocked} from "shared/state/CircuitInfo";
+
+import "./index.scss";
+
+export const LockToggleButton = () => {
```
TGCrystal, line 6: nit: 2 newlines between last import and start of code

src/site/shared/containers/Header/Left/SideBar.tsx
```diff
@@ -0,0 +1,15 @@
+import {useSharedDispatch} from "shared/utils/hooks/useShared";
+import {ToggleSideNav} from "shared/state/SideNav";
+
+import "./index.scss";
+
+export const SideBarToggleButton = () => {
```
TGCrystal, line 6: nit: 2 newlines between last import and start of code

src/site/shared/containers/Header/Left/Locker.tsx
```diff
@@ -0,0 +1,24 @@
...
+    const {isLocked} = useSharedSelector(
+        state => ({ ...state.circuit })
+    );
+    const dispatch = useSharedDispatch();
+
+	return (
+		<div>
```
TGCrystal, line 13: nit: Move this `<div>` one indentation to the right

src/site/shared/containers/Header/Left/SideBar.tsx
```diff
@@ -0,0 +1,15 @@
...
+    const dispatch = useSharedDispatch();
+
+	return (
+		<div>
```
TGCrystal, line 10: nit: one indentation to the left

src/site/shared/containers/Header/Left/SideBar.tsx
```diff
@@ -0,0 +1,15 @@
...
+            <span title="Side Bar" role="button" tabIndex={0}
+                  onClick={() => dispatch(ToggleSideNav())}>&#9776;</span>
+        </div>
```
TGCrystal, line 13: nit: one indentation to the right

TGCrystal (changes requested):
You will need to adjust the breakpoint for the dropdown arrow in the top right that appears on mobile. As it is now, the top right dropdown buttons disappear at 700px but the arrow dropdown of dropdowns doesn't appear until 600px.

LeonMontealegre (approved):

PR #997, opened 2022-02-12
Fixed bottom of canvas from getting cut off

### LINKED ISSUE #810: Bottom of canvas is cut off
The very bottom of the canvas is missing a few pixels.
image

alstar555:
Added 4 pixels to the bottom of the canvas to keep it from getting cut off. 
Fixes #810 "Bottom of canvas is cut off".

### DIFF
--- a/src/site/pages/digital/src/containers/MainDesigner/index.tsx
+++ b/src/site/pages/digital/src/containers/MainDesigner/index.tsx
@@ -108,7 +108,7 @@ export const MainDesigner = ({info, canvas}: Props) => {
                    }}>
             <canvas className="main__canvas"
                     width={w}
-                    height={h-HEADER_HEIGHT} />
+                    height={h-HEADER_HEIGHT+4} />
         </Droppable>
     </>);
 }


TGCrystal (approved):

src/site/pages/digital/src/containers/MainDesigner/index.tsx
```diff
@@ -108,7 +108,7 @@ export const MainDesigner = ({info, canvas}: Props) => {
                    }}>
             <canvas className="main__canvas"
                     width={w}
-                    height={h-HEADER_HEIGHT} />
+                    height={h-HEADER_HEIGHT+4} />
```
LeonMontealegre, line 111: Unfortunately, while this *does* fix the issue, it's more important to find out where these 4 pixels come from and make it a constant instead of hard-coding it here. 

I can almost guarantee that now every interaction with the circuit will be 4 pixels off because you also need to update the `camera` size on line 44 of this file, so we do need to be careful with this type of stuff

PR #1001, opened 2022-02-15
Make GroupActions collapsible

### LINKED ISSUE #975: History Box: Allow GroupActions to be collapsable
The history box can get cluttered pretty easily with group actions, and thus they should be collapsible.

Ideally, some sort of chevron (like `<`) would also help to indicate to the user whether or not the action is collapsed.

However this is implemented, it should be general enough so that it can be used to collapse details of other actions that would be displayed in #976

jssunray34:
I added a state variable to track whether each GroupAction should be collapsed or not. I also added a chevron button to modify this state when clicked.

Fixes #975

### DIFF
--- a/src/site/shared/containers/HistoryBox/index.tsx
+++ b/src/site/shared/containers/HistoryBox/index.tsx
@@ -9,6 +9,7 @@ import {useHistory} from "shared/utils/hooks/useHistory";
 import {CloseHistoryBox} from "shared/state/ItemNav";
 
 import "./index.scss";
+import { useState } from "react";
 
 
 type HistoryEntryProps = {
@@ -25,6 +26,8 @@ type GroupActionEntryProps = {
     g: GroupAction;
 }
 const GroupActionEntry = ({g}: GroupActionEntryProps) => {
+    const [isCollapsed, setIsCollapsed] = useState(false);
+
     if (g.isEmpty())
         return null;
     if (g.getActions().length === 1) {
@@ -33,7 +36,8 @@ const GroupActionEntry = ({g}: GroupActionEntryProps) => {
     return (
         <div className="historybox__groupentry">
             <span>Group Action</span>
-            {g.getActions().map((a, i) => {
+            <span className="historybox__groupentrycollapse" onClick={() => setIsCollapsed(!isCollapsed)}>&rsaquo;</span>
+            {isCollapsed && g.getActions().map((a, i) => {
                 return(<HistoryEntry key={`group-action-entry-${i}`} a={a}></HistoryEntry>);
             })}
         </div>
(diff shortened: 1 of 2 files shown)


TGCrystal (changes requested):
Aside from the one minor code nitpick, I think this could be merged now. However, I think there are a few small changes that should be made to improve it a little further. These changes are not set in stone either, you should run them by @LeonMontealegre as well. (Edit: I checked with Leon, he said these changes would all make sense)

- The chevron should be on the right side, like so:
  - Untitled
- The chevron should be slightly larger so that its height is that of other letters in the text like the `r`, `o`, and `u` of `Group`
- Clicking anywhere in the `GroupAction ›` "row" should apply the collapse/expansion, not just the `>`. I think that is more intuitive, and would also make tapping it easier on mobile. Note that when the action is expanded, it should **not** be closed when any of the subactions are pressed. For example, in the following screenshot, clicking any of the `Placed Clock` actions should not collapse the `GroupAction`.
  - image
- The chevron should rotate downwards when the action is expanded. If you want an example on how to do this, see the code for the button in the itemnav that transforms between the button to close and the button to open it. Specifically,
  - https://github.com/OpenCircuits/OpenCircuits/blob/master/src/site/shared/containers/ItemNav/index.tsx#L212
  - https://github.com/OpenCircuits/OpenCircuits/blob/master/src/site/shared/containers/ItemNav/index.scss#L169-L200

These are all things I probably should have put in the initial issue, sorry to dump them on you after the fact.

src/site/shared/containers/HistoryBox/index.tsx
```diff
@@ -33,7 +36,8 @@ const GroupActionEntry = ({g}: GroupActionEntryProps) => {
     return (
         <div className="historybox__groupentry">
             <span>Group Action</span>
-            {g.getActions().map((a, i) => {
+            <span className="historybox__groupentrycollapse" onClick={() => setIsCollapsed(!isCollapsed)}>&rsaquo;</span>
```
LeonMontealegre, line 39: nit: call it `historybox__groupentry-collapse` (and edit the corresponding SCSS name as well)

jssunray34:
I was not able to find a very good downward-facing chevron, so let me know if you'd like me to change this icon or its positioning. Also, I believe the chevron is the same font size as the adjacent text, but I can make it larger if needed.

TGCrystal:
You can actually use the same chevron and the `transform: rotate(90deg);` css property to rotate it to point downwards. You can have that property be part of a class that is only applied when your state variable `isCollapsed` is false. Combined with the `transition: 0.5s;` property, it will even automatically perform a smooth animation while rotating down. This is how the button to open the itemnav rotates and moves from a plus sign to an X to close it. My original comment has links that should take you right to the relevant parts of the itemnav code. If you need any further help with figuring it out, don't hesitate to let me know!

As for the font size, you can leave that as is. It looks good now that it isn't right next to the text.

jssunray34:
Oh got it! Sorry I didn't look very closely at the links you provided.

TGCrystal (approved):

LeonMontealegre (approved):

PR #1003, opened 2022-02-15
Disable Context Menu Actions if they would do nothing

### LINKED ISSUE #991: Context Menu buttons should be disabled when non-functional
The context menu (right-click menu) has buttons like "Copy", "Paste", "Cut", "Select All", "Undo", "Redo".

Undo and Redo currently are 'disabled' when there is no history in that direction.

A similar effect should be applied to:
- Copy: When there is nothing selected
- Cut: When there is nothing selected
- Select All: When there is nothing on the canvas
- Paste: ~~When there is nothing in the clipboard~~ [scrapped: has to happen asynchronously and that could cause issues]

shadowluigi27:
Disables the Cut, Copy, and Select All actions in the context menu if doing them doesn't make sense (EX: Copying nothing)

Resolves #991

### DIFF
--- a/src/site/shared/containers/ContextMenu/index.tsx
+++ b/src/site/shared/containers/ContextMenu/index.tsx
@@ -147,10 +147,10 @@ export const ContextMenu = ({info, paste}: Props) => {
                  top: `${pos?.y + HEADER_HEIGHT - CONTEXT_MENU_VERT_OFFSET}px`,
                  visibility: (isOpen ? "initial" : "hidden")
              }}>
-            <button title="Cut"        onClick={() => doFunc(onCut)}disabled = {selections.amount() === 0}>Cut</button>
-            <button title="Copy"       onClick={() => doFunc(onCopy)}disabled= {selections.amount() === 0}>Copy</button>
+            <button title="Cut"        onClick={() => doFunc(onCut)} disabled={selections.amount() === 0}>Cut</button>
+            <button title="Copy"       onClick={() => doFunc(onCopy)} disabled={selections.amount() === 0}>Copy</button>
             <button title="Paste"      onClick={() => doFunc(onPaste)}>Paste</button>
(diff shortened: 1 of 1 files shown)


src/site/shared/containers/ContextMenu/index.tsx
```diff
@@ -147,10 +147,10 @@ export const ContextMenu = ({info, paste}: Props) => {
                  top: `${pos?.y + HEADER_HEIGHT - CONTEXT_MENU_VERT_OFFSET}px`,
                  visibility: (isOpen ? "initial" : "hidden")
              }}>
-            <button title="Cut"        onClick={() => doFunc(onCut)}>Cut</button>
-            <button title="Copy"       onClick={() => doFunc(onCopy)}>Copy</button>
+            <button title="Cut"        onClick={() => doFunc(onCut)}disabled = {selections.amount() === 0}>Cut</button>
+            <button title="Copy"       onClick={() => doFunc(onCopy)}disabled= {selections.amount() === 0}>Copy</button>
```
TGCrystal, line 151: Add a space between the closing brace (`{`) of `onClick` and `disabled`

src/site/shared/containers/ContextMenu/index.tsx
```diff
@@ -147,10 +147,10 @@ export const ContextMenu = ({info, paste}: Props) => {
...
             <button title="Paste"      onClick={() => doFunc(onPaste)}>Paste</button>
-            <button title="Select All" onClick={() => doFunc(onSelectAll)}>Select All</button>
+            <button title="Select All" onClick={() => doFunc(onSelectAll)} disabled = {designer.getObjects().length === 0}>Select All</button>
```
TGCrystal, line 153: Remove spaces around the `=` that is directly after `disabled`

TGCrystal (changes requested):

TGCrystal (approved):

LeonMontealegre (approved):

PR #1006, opened 2022-02-18
Disables Utility Buttons and Popups while circuit is locked

### LINKED ISSUE #978: Expression parser still useable when circuit is locked
**Describe the bug**
The lock button in the top left of opencircuits prevents modifying the circuit in any way. This can be bypassed with the expression parser, as there is no check to prevent using it while the circuit is locked.

**Expected behavior**
When the circuit is locked, the expression parser should not be useable. There is room for discussion about whether that means you can't click Generate in the popup, the popup doesn't open, or the button to open the popup does not exist at all. It would probably make sense to remove the button to open the popup since that matches with the behavior to remove the button to open the itemnav.

lorenzoar21:
This fix is intended as a general fix for all Utilities, current and future, to be disabled while the circuit is locked. This would have future design implications that all utilities are intended for use in circuit design and modification, and I'm not sure if that's intended. If this is intended future design, then a better solution would be to disable/hide the Utilities dropdown altogether while the circuit is locked. 

This fix does not include any visual UI indications that the button is disabled due to the circuit being locked.

Fixes #978


TGCrystal (approved):

src/site/shared/containers/Header/Right/UtilitiesDropdown.tsx
```diff
@@ -27,7 +29,8 @@ export const UtilitiesDropdown = ({ extraUtilities }: Props) => {
               onClose={() => dispatch(CloseHeaderMenus())}
               btnInfo={{title: "Utilities", src: "img/icons/utilities.svg"}}>
         {extraUtilities.map(utility => (
-            <div key={utility.popupName} onClick={() => { dispatch(CloseHeaderMenus()); dispatch(OpenHeaderPopup(utility.popupName)); }}>
+            <div key={utility.popupName} onClick={() => {!info.locked && dispatch(CloseHeaderMenus()); 
+            !info.locked && dispatch(OpenHeaderPopup(utility.popupName)); }}>
```
LeonMontealegre, line 33: nit: formatting, should expand it out since the expression is a bit more complex
```tsx
<div key={utility.popupName} 
     onClick={() => {
         if (info.locked)
             return;
         dispatch(CloseHeaderMenus()); 
         dispatch(OpenHeaderPopup(utility.popupName)); 
     }}>
```

LeonMontealegre (changes requested):
Looking good!

Those are great points you brought up but I think it's fine for now and can be re-addressed when more "Utilities" are added.

However, we do definitely want a visual indicator, preferably that the Menu button just looks greyed out like the "General Tour" button in the Help Menu.

Theoretically, this shouldn't involve any new CSS since an existing CSS class should already exist and is used in the other menus

LeonMontealegre (dismissed):

Issue #971, opened 2022-02-07
History Box Improvements


TGCrystal:
There are various improvements that can be made to the history box. This issue will act like a hub, since different improvements can be implemented simultaneously by different people. Other ideas can be added to the list here.

- [ ] History Box: Make the history box resizable and moveable
- [ ] History Box: Dynamically adjust spacing to account for word wrapping
- [ ] History Box: Give custom names to GroupActions
- [ ] History Box: Allow GroupActions to be collapsable
- [ ] History Box: Allow optionally displayed extra information

PR #1016, opened 2022-02-27
Make History Box Draggable/Resizable

### LINKED ISSUE #972: History Box: Make the history box resizable and moveable
When the history box is created, it is set to a fixed size and location. Ideally, it should be both moveable and resizable. That way, the user can choose their own preferred location and size for the box.

How exactly this is implemented is open to discussion, although it should work on both desktop and mobile. A straightforward implementation that comes to mind is that it would work like a window on a computer: drag it around using the top of the window and use the corners to resize.

There should also be a button to reset the position/size of the history box, although if this proves to expand the scope/complexity of the pr too greatly, then it can be saved for a separate issue.

EricW123:
Maybe I'll continue work on other issues mentioned in #971 , so they'll all be posted in this branch
Now history box is draggable and resizzable.
Fixes #972


TGCrystal (changes requested):
Overall, this is looking great so far! I did notice a few technical bugs and a few things for further design consideration.

- The main issue is this does not seem to work on mobile. Tested in Safari on iOS 15, dragging the header does not move the history box. I was not able to test resizing because the bottom of the history box was cutoff.
- When in portrait mode (the itemnav is on the bottom) the height of the history box is locked and cannot be resized.


As for design considerations, it is probably best to consult @LeonMontealegre on these to make sure he agrees/approves:

- Comparing on Chrome (left) vs. Firefox (right) in the screenshot below, there are some visual discrepancies. I think the look on Chrome for the scrollbar is better (rounded corners, no buttons on top/bottom). As for the resize icon, the grey and larger version of Firefox looks better. Making these things consistent across browsers can range from annoying to nearly impossible, so if completely standardizing those isn't possible, that is fine.
  - image
- I think you should be able to resize using any corner/edge of the history box, similar to how you would with a window on a computer.
- Currently, the history box can cover up the header and itemnav. I don't think it should be able to cover up the header, I'm not sure if it should cover up the itemnav.
- While not a requirement, I think a button to reset the size/position of the history box could be useful. There is space for it in the history box header between the word History and the X to close it, but I'm not sure what icon would look good there, if any. Another possibility is resetting the position/size every time the history box is opened, but I think the current behavior of keeping the last used position/size makes more sense.

One last thing: Generally, it is preferred that each pr does one "thing". Thus, I think it would be preferred for history box changes to be on separate branches. Since there is one parent issue for them, it could be argued that they could be grouped together. You should ask @LeonMontealegre about this.

TGCrystal (changes requested):
Code looks good, but this is still not working on mobile (iOS) for me. Trying to drag/resize the box does nothing. Absolutely loving it on computer though, this is such a neat feature!

TGCrystal (changes requested):
Looks like it works on mobile, just one last little nitpick. As shown in the screenshot below, when the History box opens on mobile, it appears horizontally offset (and partially offscreen). I think this is due to the horizontal placement offset that would occur if the history box is first opened when the itemnav is on the left. There should be a small check of some sort so that this horizontal offset is not applied when the itemnav is on the bottom.

image

LeonMontealegre (dismissed):

TGCrystal (approved):

PR #1018, opened 2022-02-28
Update PanTool so Works with Multiple Clicks and Drags

### LINKED ISSUE #883: holding alt to pan only works for first click and drag
**Describe the bug**
While holding the alt key, the first pan via mouse works normally, but the subsequent ones don't work.

**To Reproduce**
1. Go to https://nightly.opencircuits.io/
2. Press and hold the alt key
3. Click the left mouse button and drag to pan screen
4. Release the left mouse button, without releasing the alt key
5. Click the left mouse button again and drag

**Expected behavior**
Expected behavior should be that the subsequent click and drags should pan the canvas.

 - OS: Windows
 - Browser: Chrome

riedee:
Resolves #883 

Change so you can continuing using PanTool until alt key is released.


TGCrystal (commented):
I was unable to recreate the original issue on my desktop, but I was able to on my laptop with a trackpad. This change seems to fix that issue, but it does slightly change another behavior. On nightly, if you are dragging with Alt + left click and release Alt without releasing the left click, it continues to pan until you release left click. With this change, releasing Alt switches from panning to creating a selection box. I'm not sure what the intended behavior is here, but it should be noted that it is different now.

I also was unable to test to make sure everything still works on mobile, since even without this change mobile doesn't seem to work quite right for me.

LeonMontealegre:
Okay, I think the desired behavior is the following:

Pressing ALT enters Pan mode, at this point, the PanTool is activated and if the user clicks and drags it should pan the screen.
If the user lets go of the mouse that's fine, as long as ALT is being held, the PanTool should remain active. That's what this PR fixed.

The other thing that we now want is if the user releases the ALT key while the user is still dragging, the PanTool should remain activated until the user stops dragging.

Bascially this will require `isDragging` to be set to false in `onEvent` when a `mouseup` event happens and then the deactivation condition might be as simple as `!isDragging && !input.isAltKeyDown()` (and then all the arrow key condition)

This would be a good time to write some tests!! @riedee , actually yes this should be done first to properly capture the behavior and then you can try to fix it

Lmk if you have any questions about the intended behavior

LeonMontealegre (approved):

TGCrystal (approved):

PR #1020, opened 2022-03-01
Changed BusAction to connect ports by position along an axis

### LINKED ISSUE #882: Bus feature matching inputs/outputs in wrong order
**Describe the bug**
The bus feature sometimes does not match outputs to inputs in the expected order.

**To Reproduce**
1. Place a Constant Number and a BCD Display and set their position to one of the pairs below
2. Select the outputs and inputs, click "Bus"

Pair 1: Constant Number at (-2.5,-0.17), BCD Display at (-0.4,-0.31)
Pair 2: Constant Number at (0.94,1.59), BCD Display at (2.94,1.76)

**Expected behavior**
Wires should be connected to match outputs to inputs in top to bottom order.

**Screenshots**
Here is a video demonstrating the issue:
https://user-images.githubusercontent.com/13438573/141065692-28cffa8a-0930-4ca6-8b54-8f28829e8e1b.mp4

**Desktop:**
 - OS: Ubuntu 20.04, Kernel 5.11.0, 64 bit
 - Browser: Firefox
 - Version: 94.0

tyler99b:
Ports will initially be sorted by y position unless the difference is small in which it will sort by x position (this is to handle horizontal components)
Fixes #882

### DIFF
--- a/src/app/digital/actions/addition/BusActionFactory.ts
+++ b/src/app/digital/actions/addition/BusActionFactory.ts
@@ -3,6 +3,9 @@ import {ConnectionAction} from "core/actions/addition/ConnectionAction";
 
 import {InputPort} from "digital/models/ports/InputPort";
 import {OutputPort} from "digital/models/ports/OutputPort";
+import { Port } from "core/models";
+import { GRID_SIZE } from "core/utils/Constants";
+
 
 export function CreateBusAction(outputPorts: OutputPort[], inputPorts: InputPort[]): GroupAction {
     if (inputPorts.length !== outputPorts.length)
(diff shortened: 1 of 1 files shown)


TGCrystal (approved):

LeonMontealegre (approved):

LeonMontealegre (commented):
Screen Shot 2022-03-18 at 1 33 38 PM
Actually I found a case where the behavior still isn't ideal
This happens when the ports are perpendicularly align, it's pretty much in the exact opposite order, but this is a coincidence and it's correct when the BCD Display is on the LHS of the constant number input.
I'm okay with leaving that for another issue since this is still a remarkable improvement, but if a quick-fix is doable then that would be great in this PR
@tyler99b

tyler99b:
Still needs some formatting fixes just wanted to commit the new algorithm

LeonMontealegre (approved):

PR #1022, opened 2022-03-02
Fixed Wire Delete for issue 902

### LINKED ISSUE #902: Deleting only pieces of wires
A single wire can be connected to one output port, but potentially many input ports (via splitting wires). Currently, if any single segment of the wire is selected and then deleted, the entire wire disappears. I would suggest only deleting portions of the wire that can no longer access the input port.

To understand which portions of the wire exactly can be removed, consider the fact that the the whole wire can be envisioned as a tree, with the output port as the root, intermediate nodes as nodes within the tree, and input ports as the leaf nodes.

An example scenario:
Screenshot (102)

What I would suggest should happen if I delete the selected wire:
Screenshot (103)

What actually happens:
Screenshot (104)

vhada7:
Added a boolean variable into the getPath function so that when one or more wires are deleted it only deletes the wires that are connected to those wires instead of all the wires. I also added this to the component function so that if a WirePort is deleted it doesn't delete everything.

Fixes #902


src/app/core/utils/ComponentUtils.ts
```diff
@@ -158,15 +168,17 @@ export function GetAllPaths(obj: Component): Array<Wire | Node> {
  * @param  objects The list of objects
  * @return         A SeparatedComponentCollection of the objects
  */
-export function GatherGroup(objects: IOObject[]): IOObjectSet {
+export function GatherGroup(objects: IOObject[], full: boolean = true): IOObjectSet {
+    
     const group = new IOObjectSet(objects);
-
+
```
TGCrystal, line 174: nit: remove added indentation

src/app/core/utils/ComponentUtils.ts
```diff
@@ -108,7 +108,7 @@ export function CreateGroup(objects: IOObject[]): IOObjectSet {
  * @param  w The wire to start from
  * @return   The array of wires/WirePorts in this path (including w)
```
TGCrystal, line 109: Please add a short description of the `full` parameter to these javadoc comments

src/app/core/utils/ComponentUtils.ts
```diff
@@ -158,15 +168,17 @@ export function GetAllPaths(obj: Component): Array<Wire | Node> {
  * @param  objects The list of objects
  * @return         A SeparatedComponentCollection of the objects
```
TGCrystal, line 169: Please add a short description of the `full` parameter to these javadoc comments

src/app/core/tools/handlers/DeleteHandler.ts
```diff
@@ -5,24 +5,27 @@ import {GroupAction} from "core/actions/GroupAction";
...
         (event.type === "keydown" &&
          (event.key === "Delete" || event.key === "Backspace") &&
          selections.amount() > 0),
-
     getResponse: ({history, designer, selections}: CircuitInfo) => {
-        const objs = selections.get().filter(o => o instanceof IOObject) as IOObject[];
-
+        selections.get().forEach(element => console.log(element));
+        const selectedWires = selections.get().filter(k => k instanceof Wire) as Wire[];
+        console.log(selectedWires);
```
TGCrystal, line 22: nit: remove this `console.log`

TGCrystal (changes requested):

LeonMontealegre:
Note that this now has an issue if you delete both of the endpoints of the wire-tree:
Screen Shot 2022-03-18 at 1 42 50 PM
Deleting leads to:
Screen Shot 2022-03-18 at 1 42 58 PM

I would **expect** this:
Screen Shot 2022-03-18 at 1 43 05 PM

Similarly, this happens if you delete all of the LEDs, it leaves a single node point

LeonMontealegre (approved):

TGCrystal (approved):

PR #1026, opened 2022-03-16
Improve IC Viewer

### LINKED ISSUE #754: When viewing ICs, make starting state of inputs correspond to the overall circuit
Currently, when a user chooses the View IC option, all switches and buttons default to being in the unpowered state (but can then be interacted with) (Edit: apparently they actually start in the state they were in when the IC was first created, which is I think the worst option). I believe it would be more useful if these switches/buttons represented the actual state of the IC, appearing in the powered state if their corresponding ports in the IC on the main canvas are being powered.

Note that for buttons, they would appear to be held down in the on state so long as their port on the IC is receiving power (although it would be odd for a button port to be powered, as typically the outside circuit would also have a button feeding into that port, which could not be pressed while the user views the IC).

As an example, if the overall circuit looks like this:
Screenshot (47)

Choosing to View IC on the Xor IC should look like this:
Screenshot (48)

If the user does not interact with the IC but only observes it, the state of the levers/buttons should change as their corresponding ports become powered/unpowered.

If the user _does_ interact with the buttons/switches, the IC should switch to an "editing mode" or "interactive mode", which should be in some way indicated (perhaps a color tint around the components), and with an option to revert to viewing mode, which will make the switches/buttons follow the state of the overall circuit again (and remove any color tint). While in editing mode, any changes to the state of the IC's ports in the overall circuit should _not_ change the state of the viewed buttons/switches.

If the user clicks on a button that is currently being held down, it will stay held down until the user releases the mouse, and then becomes unpowered (and for the rest of the user's time in editing mode, acts as a normal button). Whether all buttons should become released simultaneously when editing mode is entered, or just those interacted with, is up for debate (although I vote only those interacted with).

If #752 is implemented: while in viewing mode, pausing/stepping time should do so for the whole of the circuit (outside the IC included). However if the user is in editing mode, I think the entirety of the circuit outside the IC should be paused, and the user's pausing/unpausing/stepping from within the IC should have no affect on the outside circuit (although this is again up for debate, maybe the outside should remain unpaused. Either way, inside and outside should remain separate (not pausing/stepping both at the same time, as while in the IC we can't even observe the circuit outside the IC, so the user shouldn't want to step it simultaneously)).

mercuryhg31:
IC viewer now keeps IC inputs up to date with how they are on the main designer, as well as allows the user to manually interact with inputs and return to the updated view

Closes #754


src/site/pages/digital/src/containers/ICViewer/index.tsx
```diff
@@ -66,7 +73,13 @@ export const ICViewer = (() => {
             // Add input listener
             info.input.addListener((event) => {
                 const change = toolManager.onEvent(event, info);
-                if (change) renderer.render();
+                if (change) {
+                    renderer.render();
+                    const worldMousePos = camera.getWorldPos(info.input.getMousePos());
+                    const obj = FindPressableObjects(worldMousePos, info);
+                    if (isPressable(obj) && event.type === "click")
```
mercuryhg31, line 80: Left this as `"click"` only since `"mousedown"` and `"mouseup"` would cause `pauseUpdates` to trigger if you just dragged your mouse across an object. I think it's alright if the callback doesn't pause right away if you press mouse down on a `Pressable` since it wouldn't actually toggle until you also released (or vice versa, basically just need an actual click event to toggle value and have reason for pausing).

LeonMontealegre: Hmmm let's talk about this in class, because how does this work with a Button?

src/app/core/tools/InteractionTool.ts
```diff
@@ -5,29 +5,17 @@ import {Event} from "core/utils/Events";
 import {isPressable} from "core/utils/Pressable";
 import {LEFT_MOUSE_BUTTON} from "core/utils/Constants";
 
-import {CircuitDesigner, IOObject} from "core/models";
-
 import {DefaultTool} from "./DefaultTool";
 
 import {EventHandler} from "./EventHandler";
+import {FindPressableObjects} from "core/utils/ComponentUtils";
```
LeonMontealegre, line 11: nit: move this to the rest of `core/utils` imports above (see https://docs.opencircuits.io/Guides/Imports)

src/app/core/tools/InteractionTool.ts
```diff
@@ -5,29 +5,17 @@ import {Event} from "core/utils/Events";
 import {isPressable} from "core/utils/Pressable";
 import {LEFT_MOUSE_BUTTON} from "core/utils/Constants";
```
LeonMontealegre, line 6: nit: this should be very top import

src/app/core/utils/ComponentUtils.ts
```diff
@@ -20,10 +18,7 @@ import {Component} from "core/models/Component";
 import {Wire} from "core/models/Wire";
 import {Node, isNode} from "core/models/Node";
 import {Port} from "core/models/ports/Port";
-import {ConnectionAction, DisconnectAction} from "core/actions/addition/ConnectionAction";
-import {CircuitDesigner} from "core/models/CircuitDesigner";
-import {DeleteAction, PlaceAction} from "core/actions/addition/PlaceAction";
-import {TranslateAction} from "core/actions/transform/TranslateAction";
+import {CircuitInfo} from "core/utils/CircuitInfo";
```
LeonMontealegre, line 21: nit: import order

src/app/core/utils/ComponentUtils.ts
```diff
@@ -151,6 +146,23 @@ export function GetAllPaths(obj: Component): Array<Wire | Node> {
...
 
+/**
+ * Very specifically get the objects and wires and reverse them SEPARATELY.
+ *  Doing `designer.getAll().reverse()` would put the wires BEFORE the objects
+ *  which will cause incorrect behavior! Objects are always going to need to be
+ *  pressed/selected before wires!
+ * 
+ * @param  pos   Position on designer
+ * @param  info  Uses designer from CircuitInfo
+ * @returns      Array of pressable objects at pos
```
LeonMontealegre, line 157: It returns only the first pressable object at pos

src/app/digital/models/ioobjects/other/IC.ts
```diff
@@ -97,7 +97,7 @@ export class IC extends DigitalComponent {
     }
 
     public getData(): ICData {
-        return this.data;
+        return ICData.Create(this.collection)!;
```
LeonMontealegre, line 100: What was the reason for this again?
It scares me greatly
If you need to copy the ICData at some point, the copying should probably be done in the ICViewer somehow

src/site/pages/digital/src/containers/ICViewer/index.tsx
```diff
@@ -24,6 +24,9 @@ import {GetRenderFunc} from "site/digital/utils/Rendering";
 import {CloseICViewer} from "site/digital/state/ICViewer";
 
 import "./index.scss";
+import {Button, Switch} from "digital/models/ioobjects/inputs";
+import {IC} from "digital/models/ioobjects";
+import {isPressable} from "core/utils/Pressable";
```
LeonMontealegre, line 29: nit: import order

src/site/pages/digital/src/containers/ICViewer/index.tsx
```diff
@@ -116,6 +141,22 @@ export const ICViewer = (() => {
             dispatch(CloseICViewer());
         }
 
+        const updateViewer = () => {
+            try {
```
LeonMontealegre, line 145: Why is there a `try-catch` here?

LeonMontealegre (changes requested):

TGCrystal (changes requested):
Looks good, only issue I found is that on mobile, the Restore and Close buttons look a little compressed

image

LeonMontealegre (approved):

TGCrystal (approved):

Issue #710, opened 2021-09-26
Switch to env variables instead of flags for backend


LeonMontealegre:
Besides running locally (which is already handled by the build scripts), `nightly` and the regular site both use GCP which only allows us to pass environment variables to configure the backend server, not flags.

Seems easier to simply just switch to entirely environment variables

PR #1030, opened 2022-03-18
Switch Flags to ENV variables for backend

brandon-red:
Changed all flags in `main.go` to env variables to address #710.


src/server/main.go
```diff
@@ -32,46 +31,47 @@ func getPort() string {
 	return "8080"
 }
 
+// Check if env variable exists, if not set to default value, return value of env variable
+func checkEnv(key string, defaultVal string) string {
+	if os.Getenv(key) == "" {
+		os.Setenv(key, defaultVal)
```
KevinMackenzie, line 37: I'd say don't bother setting the env vars when using the default.  We should only be checking these on startup anyway.

src/server/main.go
```diff
@@ -2,7 +2,6 @@ package main
 
 import (
 	"context"
-	"flag"
```
KevinMackenzie, line 5: Not mandatory, but it would be cool to have a `flag`-like interface that reads from env vars instead so we keep types.

brandon-red: Other than `bool` or `string`, are there other types that will need to be considered in the future? For now, the change I am implementing will handle only those two, but can be expanded on if need be.

KevinMackenzie: `int` and `uint` if you're up for it.  We probably don't care about differentiating between the different sizes (like `uint16` and `uint32`)

src/server/main.go
```diff
@@ -32,48 +33,59 @@ func getPort() string {
...
-	googleAuthConfig := flag.String("google_auth", "", "<path-to-config>; Enables google sign-in API login")
-	noAuthConfig := flag.Bool("no_auth", false, "Enables username-only authentication for testing and development")
-	userCsifConfig := flag.String("interface", "sqlite", "The storage interface")
-	sqlitePathConfig := flag.String("sqlitePath", "sql/sqlite", "The path to the sqlite working directory")
-	dsEmulatorHost := flag.String("ds_emu_host", "", "The emulator host address for cloud datastore")
-	dsProjectId := flag.String("ds_emu_project_id", "", "The gcp project id for the datastore emulator")
-	ipAddressConfig := flag.String("ip_address", "0.0.0.0", "IP address of server")
-	portConfig := flag.String("port", "8080", "Port to serve application, use \"auto\" to select the first available port starting at 8080")
+	// Flag for noAuthConfig to keep boolean type
+	noAuthConfig := flag.Bool("no_auth", toBool(os.Getenv("no_auth")), "Enables username-only authentication for testing and development")
```
KevinMackenzie, line 57: I like this strategy.  Checking `flag`s first and fallback on the environment variables.  This should work in prod just fine and be convenient for back-end dev too.

TGCrystal:
Not required, but I think environment variables are usually in all caps, so it wouldn't hurt for these ones to be all caps as well. An `OC_` prefix or something like that could also be nice to distinguish these as OpenCircuits specific.

KevinMackenzie:
These changes got pulled into #1191

Issue #656, opened 2021-05-03
Fit to Screen is placing components at where the itemNav is located


qihenry:
**Describe the bug**
When the site is not full-screen and with at least 2 component with some distance apart, pressing f to fit to screen will cause it zoom in a way such that the left-most component will located behind the ItemNav

**To Reproduce**
1. Put OpenCircuits to not full-screen
2. Add two Components away from each other 
2. Click F
3. See that the left component is behind the ItemNav

**Expected behavior**
The expected behavior should be the same behavior as when the site is in Full Screen when F is pressed. Fit to screen should place the components away from the ItemNav when it is opened. 
**Screenshots**
![fittoscreenBug](https://user-images.githubusercontent.com/24613817/116833698-aeb60b00-ab88-11eb-8f1f-6310b6418345.gif)

**Desktop (please complete the following information):**
 - OS: Windows 10 Education
 - Browser: Brave

PR #1031, opened 2022-03-19
Ensure Focus doesn't put components behind opened Itemnav

shadowluigi27:
- Added a new Margin type to the Camera class that controls the camera margin when using the focus command (Shortcut F).  
 
- Added a new effect in the itemnav.tsx file that updates the left margin when the itemnav opens or closes

- Changed the getCameraFit utility Function to take into account the margin  (Massive help from Leon on this)

Fixes: #656 

Old Behavior

https://user-images.githubusercontent.com/72285616/159103662-d3621745-03b3-4086-af29-c31605f0bea5.mp4

New Behavior

https://user-images.githubusercontent.com/72285616/159103650-4198bedc-da5e-4aed-bc53-1c461aa19f61.mp4


src/app/core/utils/ComponentUtils.ts
```diff
@@ -321,17 +321,23 @@ export function GetCameraFit(camera: Camera, objs: CullableObject[], padding: nu
     if (objs.length === 0)
         return [V(), 1];
 
+    const MarginSize = V(camera.getMargin().left - camera.getMargin().right, camera.getMargin().top - camera.getMargin().bottom);
```
LeonMontealegre, line 324: nit: call it `marginSize`

src/app/core/utils/ComponentUtils.ts
```diff
@@ -321,17 +321,23 @@ export function GetCameraFit(camera: Camera, objs: CullableObject[], padding: nu
...
     const worldSize = camera.getWorldPos(screenSize).sub(camera.getWorldPos(V(0,0))); // World size of camera view
 
     // Determine which bbox dimension will limit zoom level
     const xRatio = bbox.getWidth()/worldSize.x;
     const yRatio = bbox.getHeight()/worldSize.y;
     const finalZoom = camera.getZoom()*Math.max(xRatio, yRatio)*padding;
 
+    // Only subtract off 0.5 of the margin offset since currently it's centered on the margin'd
+    //  screen size so only half of the margin on the top/left need to be contributed
+    const finalPos = newPos.sub(MarginSize.scale(0.5 * finalZoom));
```
LeonMontealegre, line 339: Did you try this with a different margin? like with something on the right / top / bottom? 
I feel like the math here is still slightly wrong because I think you need more than just `MarginSize`, specifically I think you need to be `.sub`ing the vector `V(camera.getMargin().left, camera.getMargin().bottom)` for calculation of `finalPos` and using `MarginSize` only for the first `.sub` for calculation of `screenSize`. Or vice-versa. Or maybe it's somehow fine but I'm skeptical.

src/app/core/utils/math/Camera.ts
```diff
@@ -6,7 +6,7 @@ import {Matrix2x3} from "./Matrix";
 import {TransformContains} from "./MathUtils";
 
 
-
+type Margin = {left: number, right: number, bottom: number, top: number}
```
LeonMontealegre, line 9: nit: but blank line after this and before the comment below it

src/app/core/utils/math/Camera.ts
```diff
@@ -143,6 +146,9 @@ export class Camera {
     public getCenter(): Vector {
         return V(this.width/2, this.height/2);
     }
+    public getSize(): Vector {
```
LeonMontealegre, line 149: Probably should add a JSDoc comment to this like the rest of the methods

src/app/core/utils/math/Camera.ts
```diff
@@ -197,4 +203,23 @@ export class Camera {
...
+
+    /**
+     * This sets the margin for the camera
+     * @param left the left margin of the camera
+     * @param right the right margin of the camera
+     * @param bottom the bottom margin of the camera
+     * @param top the top margin of the camera
+     */
+    public setMargin(newMargin: Partial<Margin>): void
+    {
```
LeonMontealegre, line 222: nit: keep `{` on same line, following conventions of the rest of the codebase

src/site/shared/containers/ItemNav/index.tsx
```diff
@@ -122,6 +122,11 @@ export const ItemNav = <D,>({ info, config, additionalData,
         return () => DragDropHandlers.removeListener(resetListener);
     }, [setState]);
 
+    //updates camera margin when itemnav is rendered open
+    useEffect( () => {
```
LeonMontealegre, line 126: nit: remove space before `()`

LeonMontealegre (changes requested):

src/app/core/utils/ComponentUtils.ts
shadowluigi27: So I tried it out with the bottom/top margin and yea, the code here doesn't work on it. The fix that seems to work is the vice versa from your suggestion.

TGCrystal (dismissed):

LeonMontealegre (approved):

PR #1038, opened 2022-03-29
Combination of create and store

### LINKED ISSUE #504: Merge the "Create" and "Update" paths for circuits
Its inconvenient to require both of these operations separately for clients of the API and it opens up some possible race conditions (not security related, but could potentially break new circuits if adversarial).

JackScofield:
Write a new handler function to cover both store and create requests. 

The new circuitUpdateHandler function basically checks for the existence of the circuit based on the id. If there is no circuit, then create a new one. Otherwise, update the circuit. Return is the same, parse the Metadata back to the client.

Fixes #504

### DIFF
--- a/src/server/api/circuits.go
+++ b/src/server/api/circuits.go
@@ -2,7 +2,6 @@ package api
 
 import (
 	"encoding/json"
-	"fmt"
 	"io"
 	"io/ioutil"
 	"net/http"
@@ -30,51 +29,6 @@ func parseCircuitRequestData(r io.Reader) (model.Circuit, error) {
 		return model.Circuit{}, err
 	}
 	return newCircuit, nil
-}
-
-func circuitStoreHandler(m interfaces.CircuitStorageInterfaceFactory, c *gin.Context, userId model.UserId) {
-	circuitId := c.Param("id")
-
-	storageInterface := m.CreateCircuitStorageInterface()
-	circuit := storageInterface.LoadCircuit(circuitId)
-	if circuit == nil {
-		c.JSON(http.StatusNotFound, nil)
-		return
-	}
-	if circuit.Metadata.Owner != userId {
-		c.JSON(http.StatusForbidden, nil)
-		return
-	}
-
-	newCircuit, err := parseCircuitRequestData(c.Request.Body)
-	if err != nil {
-		c.JSON(http.StatusBadRequest, err.Error())
-		return
-	}
-
-	circuit.Update(newCircuit)
-	storageInterface.UpdateCircuit(*circuit)
-
-	// Returned the updated metadata so the client can get any changes the server made to it
-	c.JSON(http.StatusAccepted, circuit.Metadata)
-}
-
-func circuitCreateHandler(m interfaces.CircuitStorageInterfaceFactory, c *gin.Context, userId model.UserId) {
-	storageInterface := m.CreateCircuitStorageInterface()
-
-	newCircuit, err := parseCircuitRequestData(c.Request.Body)
-	if err != nil {
-		c.JSON(http.StatusBadRequest, err.Error())
-		return
-	}
-
-	circuit := storageInterface.NewCircuit()
-	circuit.Metadata.Owner = userId
-	circuit.Update(newCircuit)
-	storageInterface.UpdateCircuit(circuit)
-
-	// Returned the updated metadata so the client can get any changes the server made to it
-	c.JSON(http.StatusAccepted, circuit.Metadata)
 }
 
 func circuitLoadHandler(m interfaces.CircuitStorageInterfaceFactory, c *gin.Context, userId model.UserId) {
@@ -116,8 +70,7 @@ func circuitUpdateHandler(m interfaces.CircuitStorageInterfaceFactory, c *gin.Co
 	} else {
 		circuit := storageInterface.LoadCircuit(circuitId)
 		if circuit == nil {
-			fmt.Println("nil for circuit")
-			fmt.Println("circuitId is", circuitId)
+			c.JSON(http.StatusNotFound, nil)
 		}
 		if circuit.Metadata.Owner != userId {
 			c.JSON(http.StatusForbidden, nil)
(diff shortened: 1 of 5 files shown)


src/server/api/circuits.go
```diff
@@ -95,6 +95,43 @@ func circuitLoadHandler(m interfaces.CircuitStorageInterfaceFactory, c *gin.Cont
...
+	circuit := storageInterface.LoadCircuit(circuitId)
+	if circuit == nil {
+		needToCreate = true
+	}
+	newCircuit, err := parseCircuitRequestData(c.Request.Body)
+	if err != nil {
+		c.JSON(http.StatusBadRequest, err.Error())
+		return
+	}
+	if needToCreate {
```
KevinMackenzie, line 113:
```suggestion
	if circuit := storageInterface.LoadCircuit(circuitId); circuit != nil {
```
If you flip the conditional you can use this neat syntax feature.  We probably should be parsing the request body before loading the circuit anyway too.

KevinMackenzie (commented):
Is it feasible to delete the old routes and update the front-end to use only this new one?  That's the other part that needs to be done before #504 can be closed.

src/server/api/circuits.go
JackScofield: Yes, that's why I didn't write Fixes #504 at the end. I will refactor the code and maybe commit tomorrow.

JackScofield:
Just refractor the code for both front-end and back-end. All the user operations are not changed. Delete create function, and the creation and changing of circuits use Update function.

src/site/shared/api/Circuits.ts
```diff
@@ -2,23 +2,23 @@ import {CircuitMetadata, CircuitMetadataDef} from "core/models/CircuitMetadata";
...
+//         headers: {
+//             "authType": auth.getType(),
+//             "authId": auth.getId()
+//         },
+//         data
+//     })) as CircuitMetadataDef);
+// }
 
-export async function CreateUserCircuit(auth: AuthState, data: string): Promise<CircuitMetadata | undefined> {
+export async function UpdateUserCircuit(auth: AuthState, data: string, circuitId: string): Promise<CircuitMetadata | undefined> {
```
KevinMackenzie, line 18: The `circuitId` here feels like the right place to use a nullable string rather than rely on empty string as a special value.  Is that idiomatic @LeonMontealegre.  It might have other implications though.

src/site/shared/api/Circuits.ts
```diff
@@ -2,23 +2,23 @@ import {CircuitMetadata, CircuitMetadataDef} from "core/models/CircuitMetadata";
...
+//     return new CircuitMetadata(JSON.parse(await Request({
+//         method: "POST",
+//         url: "api/circuits",
+//         headers: {
+//             "authType": auth.getType(),
+//             "authId": auth.getId()
+//         },
+//         data
+//     })) as CircuitMetadataDef);
+// }
```
KevinMackenzie, line 16: Just delete this.  The git history will preserve it if it's needed again.

src/server/api/circuits.go
```diff
@@ -95,6 +96,61 @@ func circuitLoadHandler(m interfaces.CircuitStorageInterfaceFactory, c *gin.Cont
...
+		circuit := storageInterface.NewCircuit()
+		circuit.Metadata.Owner = userId
+		circuit.Update(newCircuit)
+		storageInterface.UpdateCircuit(circuit)
+		c.JSON(http.StatusAccepted, circuit.Metadata)
+	} else {
+		circuit := storageInterface.LoadCircuit(circuitId)
+		if circuit == nil {
+			fmt.Println("nil for circuit")
+			fmt.Println("circuitId is", circuitId)
```
KevinMackenzie, line 120: Switch this to a `StatusNotFound` error and don't bother printing to the log.

src/server/api/circuits.go
```diff
@@ -95,6 +96,61 @@ func circuitLoadHandler(m interfaces.CircuitStorageInterfaceFactory, c *gin.Cont
 	c.JSON(http.StatusOK, circuit)
 }
 
+func circuitUpdateHandler(m interfaces.CircuitStorageInterfaceFactory, c *gin.Context, userId model.UserId) {
```
KevinMackenzie, line 99: Can you delete the "create" and "store" handler above now that this one function replaces them?

src/server/api/circuits.go
```diff
@@ -95,6 +96,61 @@ func circuitLoadHandler(m interfaces.CircuitStorageInterfaceFactory, c *gin.Cont
...
+	//		c.JSON(http.StatusForbidden, nil)
+	//		return
+	//	}
+	//
+	//	circuit.Update(newCircuit)
+	//	storageInterface.UpdateCircuit(*circuit)
+	//
+	//	// Returned the updated metadata so the client can get any changes the server made to it
+	//	c.JSON(http.StatusAccepted, circuit.Metadata)
+	//}
```
KevinMackenzie, line 151: Delete this commented-out code too

src/server/api/circuits.go
```diff
@@ -95,6 +96,61 @@ func circuitLoadHandler(m interfaces.CircuitStorageInterfaceFactory, c *gin.Cont
...
+
+	circuitId := c.Param("id")
+
+	newCircuit, err := parseCircuitRequestData(c.Request.Body)
+	if err != nil {
+		c.JSON(http.StatusBadRequest, err.Error())
+		return
+	}
+
+	if circuitId == "0" {
```
KevinMackenzie, line 110: +1 for "0" circuit Id meaning "create".  That's probably better behavior than "circuit ID not found, let's create a new circuit".

src/site/shared/state/thunks/SaveCircuit.ts
```diff
@@ -19,12 +19,14 @@ export function SaveCircuit(data: string): ThunkResult<Promise<boolean>> {
...
-
         dispatch(_SetCircuitSavingStart());
 
         try {
-            const newData = await (id ? UpdateUserCircuit(auth!, id, data) :
-                                        CreateUserCircuit(auth!, data));
+            // const newData = await (UpdateUserCircuit(auth!, id, data));
+
+            // const newData = await (id ? UpdateUserCircuit(auth!, id, data) :
+            //     CreateUserCircuit(auth!, data));
```
KevinMackenzie, line 28: Delete this commented-out code too

KevinMackenzie (commented):
Looking good overall.  Suggestions are mostly about cleanup.

src/site/shared/api/Circuits.ts
LeonMontealegre: Agreed, yeah use `circuitId?: string` and pass undefined when it's not defined

PR #1040, opened 2022-04-01
History box now contains redo history

### LINKED ISSUE #1027: History Box: Add redo history view
The redo history should also be viewable in the History Box.

The entries should be exactly the same as the undo history except they are greyed out or have a dashed border or some other visual indicator that they are different.
They'll always be the top entries in the box since the entries must be ordered from top to bottom as most recent to least recent actions and redo actions are sort-of "in the future" in that regard.

There should also be a visual separator between the top entry of the undo history and the bottom entry of the redo history, most likely just a line going across the box.

mattsweet417:
History box does not contain redo history, so I added new dashed entry types and used them when redo history exists. Also, I added a divider between undo and redo history. 

Fixes #1027 

hbox2

hbox3


src/site/shared/containers/HistoryBox/index.tsx
```diff
@@ -15,49 +15,80 @@ import "./index.scss";
...
+                <span
+                    className={`historybox__groupentry__collapse_btn \
+                                ${isCollapsed ? "historybox__groupentry__collapse_btn-collapsed" : "" }`}>
+                    &rsaquo;
+                </span>
+                {!isCollapsed && g.getActions().map((a, i) => {
+                    return(<HistoryEntry key={`group-action-dashedentry-${i}`} a={a} isRedo={true} ></HistoryEntry>);
+                })}
+            </div>
+        );
```
TGCrystal, line 91: This can probably be condensed and just use a ternary operator for the first className and the key, since it seems like those are the only differences between the if and the else

src/site/shared/containers/HistoryBox/index.tsx
```diff
@@ -69,21 +100,48 @@ export const HistoryBox = ({ info }: Props) => {
         state => ({ ...state.itemNav })
     );
     const dispatch = useSharedDispatch();
-
+
```
TGCrystal, line 103: nit: delete this unnecessary indent

src/site/shared/containers/HistoryBox/index.tsx
```diff
@@ -69,21 +100,48 @@ export const HistoryBox = ({ info }: Props) => {
...
-
+    
     const { undoHistory, redoHistory } = useHistory(info);
 
-    return (
-        <div className={`historybox ${isOpen ? "" : "historybox__move"} ${isHistoryBoxOpen ? "" : "hide"}`}>
-            <div>
-                <span>History</span>
-                <span onClick={() => dispatch(CloseHistoryBox())}>×</span>
+    var redoHasMembers = false;
```
TGCrystal, line 106: nit: use `let` instead of `var`

TGCrystal: Actually it doesn't look like this variable is used at all, it can probably just be deleted

src/site/shared/containers/HistoryBox/index.tsx
```diff
@@ -69,21 +100,48 @@ export const HistoryBox = ({ info }: Props) => {
...
+                    <span>History</span>
+                    <span onClick={() => dispatch(CloseHistoryBox())}>×</span>
+                </div>
+                <div>
+                    <div style={{textAlign: 'center', fontWeight: 'bold'}}> Undo </div>
+                    {[...undoHistory].reverse().map((a, i) =>
+                        <HistoryEntry key={`history-box-entry-${i}`} a={a} isRedo={false}></HistoryEntry>
+                    )}
+                    
+                </div>
```
TGCrystal, line 142: Similar to the large if/else statement above, this can probably be compressed. Inside of the html sections, you can use the `{}` braces to enter back into the javascript. Bascially, you would want something like this:
```ts
{ if (redoHistory.length > 0) {<>
    {[...redoHistory].reverse().map((a, i) =>
        <HistoryEntry key={`history-box-dashedentry-${i}`} a={a} isRedo></HistoryEntry>
    )}
    <div style={{textAlign: 'center', fontWeight: 'bold'}}> Redo </div>
    <div className={"historybox__separator"} > </div>
</>}
}
```
I may have the syntax a little off there, but that reduce the amount of code reused between the if and the else, since it looks like everything else is the same.

TGCrystal (changes requested):

mattsweet417:
Thank you for the comments! I am completely new to web dev, and I had no idea how to condense it. Thanks for taking your time to explain it :).

src/site/shared/containers/HistoryBox/index.tsx
```diff
@@ -71,19 +71,26 @@ export const HistoryBox = ({ info }: Props) => {
     const dispatch = useSharedDispatch();
 
     const { undoHistory, redoHistory } = useHistory(info);
-
```
TGCrystal, line 74: nit: add this empty line back in

src/site/shared/containers/HistoryBox/index.tsx
```diff
@@ -71,19 +71,26 @@ export const HistoryBox = ({ info }: Props) => {
...
                 <span onClick={() => dispatch(CloseHistoryBox())}>×</span>
             </div>
             <div>
+                { redoHistory.length > 0 ? (<>
+                    {[...redoHistory].reverse().map((a, i) =>
+                        <HistoryEntry key={`history-box-dashedentry-${i}`} a={a} isRedo></HistoryEntry>
+                    )}
+                    <div style={{textAlign: 'center', fontWeight: 'bold'}}> Redo </div>
+                    <div className={"historybox__separator"} > </div>
+                 </>) : (<></>)}
```
TGCrystal, line 87: You can go even crazier with this and instead of doing a ternary operator where the false case is just empty, you can replace it with `&&`. Broadly, it would look like `(condition) && (html)`. Here's an example of it in place: https://github.com/OpenCircuits/OpenCircuits/blob/2746d42f1d75d5097514c08b54dd449641c3865e/src/site/shared/containers/Header/Right/index.tsx#L33-L35

src/site/shared/containers/HistoryBox/index.tsx
```diff
@@ -15,46 +15,46 @@ import "./index.scss";
...
+const HistoryEntry = ({ a, isRedo }: HistoryEntryProps) => {
     if (a instanceof GroupAction)
-        return (<GroupActionEntry g={a}></GroupActionEntry>);
-    return (
-        <div className="historybox__entry"
-             // Necessary to stop child entries from collapsing the parent history entry
-             onClick={(e) => e.stopPropagation()}>
-            {a.getName()}
+        return (<GroupActionEntry g={a} isRedo={isRedo}></GroupActionEntry>);
+    return(
```
TGCrystal, line 23: nit: add the space back between `return` and `(`

TGCrystal (changes requested):
Looking good, just a few more small changes to make!

src/site/shared/containers/HistoryBox/index.tsx
```diff
@@ -15,46 +15,46 @@ import "./index.scss";
...
-        <div className="historybox__entry"
-             // Necessary to stop child entries from collapsing the parent history entry
-             onClick={(e) => e.stopPropagation()}>
+        <div className={isRedo ? "historybox__dashedentry" : "historybox__entry"}
+            // Necessary to stop child entries from collapsing the parent history entry
+            onClick={(e) => e.stopPropagation()}>
             {a.getName()}
         </div>
-    );
+    )
```
TGCrystal, line 29: nit: add the semicolon back in after this `)`

src/site/shared/containers/HistoryBox/index.tsx
```diff
@@ -71,19 +71,27 @@ export const HistoryBox = ({ info }: Props) => {
     const dispatch = useSharedDispatch();
 
     const { undoHistory, redoHistory } = useHistory(info);
-
+
```
TGCrystal, line 74: nit: this should be an empty line, remove the indentation

TGCrystal (changes requested):

TGCrystal (dismissed):

src/site/shared/containers/HistoryBox/index.scss
```diff
@@ -121,4 +142,33 @@ $historybox-width: 250px;
         }
     }
 
+    &__dashedgroupentry {
```
LeonMontealegre, line 145: this and `&__groupentry` are nearly identical besides a few differences, what you can do is make a class `&__groupentry--dashed` and then below when you apply the className, you can have `historybox__groupentry ${isRedo ? "historybox__groupentry--dashed" : ""}` which applies both classes, it's a little verbose but is the "preferred" way of doing things; but you definitely want to avoid duplication as much as possible

LeonMontealegre: Also same thing with `&__entry` and `&__dashedentry`

LeonMontealegre (changes requested):
Looking good! Just some CSS changes

LeonMontealegre (approved):

PR #1042, opened 2022-04-05
Allows selections of multiple types of gate to change input size

### LINKED ISSUE #820: Cannot change number of inputs when selecting multiple types of gates
**Describe the bug**
When selecting one or more gates of the same type (or the gate's type and its negation (AND with NAND, OR with NOR, and XOR with XNOR)), the selection popup has an "Input Count" input field. However, this input field disappears when gates not of the same type are selected (i.e. OR and XOR), but it should still be present.

**To Reproduce**
1. Place an OR gate and an XOR gate on the canvas
2. Select both, and observe the selection popup without the Input Count field on it

**Expected behavior**
I would expect the Input Count field to appear so long as the user is only selecting some combination of: AND, NAND, OR, NOR, XOR, and XNOR gates.

 - OS: Windows 10
 - Browser: Chrome

**Additional context**
The behavior with the other components with varying input/output numbers is currently appropriate (Encoders and Decoders, one is input count and the other is output count, so not allowing both the be changed at once makes sense. And they both go down to 1, so it makes sense that they can't be done along with gates which only go down to 2. MUX and DEMUX can be changed together, which is good, and all the segment displays can be changed together as well).

shadowluigi27:
I changed the input module from specifying each type of gate, which only allowed selection of gates and their not version to be able to change input (EX: OR and NOR) but not 2 different types of gates (OR and AND). 

By generalizing the type to 'Gate', a selection of  only gates no matter the type are now allowed to change inputs when selected together.

Old Behavior
image

New Behavior
image

Fixes #820


src/site/pages/digital/src/containers/SelectionPopup/modules/InputCountModule.tsx
```diff
@@ -1,20 +1,16 @@
...
 import {InputPortChangeAction} from "digital/actions/ports/InputPortChangeAction";
 
 import {CreateModule, ModuleConfig, PopupModule} from "shared/containers/SelectionPopup/modules/Module";
 
 
-const Config: ModuleConfig<[ANDGate, NANDGate, ORGate,
-                            NORGate, XORGate, XNORGate], number> = {
-    types: [ANDGate, NANDGate, ORGate,
-            NORGate, XORGate, XNORGate],
+
```
TGCrystal, line 11: nit: only two spaces between last import and start of code, so this empty line is unnecessary

TGCrystal (changes requested):
`Gate` includes both `Buffer` and `Not`, which do not support changing the input number count. Instead of generalizing to `Gate`, you can do `ANDGate | ORGate | XORGate` (I think that would automatically take care of the inverted versions of those).

src/site/pages/digital/src/containers/SelectionPopup/modules/InputCountModule.tsx
TGCrystal: This still needs to be addressed, but looks good otherwise

TGCrystal (changes requested):

TGCrystal (approved):

src/site/shared/containers/SelectionPopup/modules/Module.tsx
```diff
@@ -98,8 +99,17 @@ export const CreateModule = (<T extends any[], P extends ModuleTypes>(props: Mod
         return v;
     }
 
+    const checkType = (s: Selectable, Type: Function & { prototype: T[number]; }) => {
+        // check if s instanceof Type and also not instnace of anything in config.exclude
+
+        if(config.exclude)
+        {for (let exclusion of config.exclude)  { if(s instanceof exclusion) {return false;}}}
```
LeonMontealegre, line 106: I think this can be condensed down to 
```ts
if (config.exclude?.some(exclusion => s instanceof exclusion))
    return false;
```

src/site/shared/containers/SelectionPopup/modules/Module.tsx
```diff
@@ -98,8 +99,17 @@ export const CreateModule = (<T extends any[], P extends ModuleTypes>(props: Mod
         return v;
     }
 
+    const checkType = (s: Selectable, Type: Function & { prototype: T[number]; }) => {
+        // check if s instanceof Type and also not instnace of anything in config.exclude
```
LeonMontealegre, line 103: nit: capitalize first letter of comment, also typo on `instance`, also mention this was necessary for issue #820

LeonMontealegre (changes requested):
Looking good, just a few formatting issues!

LeonMontealegre (approved):

PR #1044, opened 2022-04-06
Item Placing Improvement

### LINKED ISSUE #863: Selection box interferes with tap to multicreate feature
**Describe the bug**
When using the tap to place feature and dragging to create a selection at the same time, sometimes the selection action causes the component to not be placed.

**To Reproduce**
1. Go to [nightly.opencircuits.io](https://nightly.opencircuits.io/)
2. Drag a component to the canvas.
3. Tap a component in the item nav multiple times
4. Click and drag on the canvas, releasing while hovering over the previously placed component.

**Expected behavior**
The components should be placed and ideally the selection box should not be created while items are in the tap to place queue.

**Screenshots**
Example of components not being placed:
tapNotCreate
Example of components sometimes being placed:
tapCreate

riedee:
Resolves #863 

Disables the selection box tool when placing items.
Creates isPlacing const as part of ItemNav state, which is used to disable selection popup when placing items.


src/site/shared/containers/SelectionPopup/index.tsx
```diff
@@ -78,6 +82,12 @@ export function SelectionPopup({info, modules, docsUrlConfig}: Props) {
         });
     }, input, [setClickThrough]);
 
+    // If placing, deselect all selections and re-render
+    useEffect(() => {
+        if (!isPlacing)
+            history.add(CreateDeselectAllAction(info.selections).execute());
+            renderer.render();
+    }, [isPlacing])
```
TGCrystal, line 90: nit: add semicolon to the end of this line

src/site/shared/containers/ItemNav/index.tsx
```diff
@@ -122,6 +132,15 @@ export const ItemNav = <D,>({ info, config, additionalData,
...
     }, [setState]);
 
+    // Keep track of if item is being placed
+    useEffect(() => {
+        if (curItemID) {
+            setIsPlacing();
+        }
+        else {
+            setIsPlacing();
+        }
```
TGCrystal, line 142: I may be missing something, but is this if else statement necessary? It is doing the same thing regardless of the condition

TGCrystal (changes requested):

src/site/shared/containers/ItemNav/index.tsx
riedee: No that does not do anything, I definitely just missed it before committing; it should be fixed now!

src/site/shared/containers/SelectionPopup/index.tsx
```diff
@@ -78,6 +82,12 @@ export function SelectionPopup({info, modules, docsUrlConfig}: Props) {
         });
     }, input, [setClickThrough]);
 
+    // If placing, deselect all selections and re-render
+    useEffect(() => {
+        if (!isPlacing)
+            history.add(CreateDeselectAllAction(info.selections).execute());
+            renderer.render();
+    }, [isPlacing]);
```
LeonMontealegre, line 90: I forget if we talked about it -- but I think we decided that deselecting everything was probably not the way to go about this? I believe the SelectionPopup actually has a state variable `clickThrough` that allows mouse events to go straight thru the SelectionPopup, so that's probably what we want? Or we would just want to not show the SelectionPopup while `isPlacing` is true.

src/site/shared/containers/ItemNav/index.tsx
```diff
@@ -83,8 +83,11 @@ export const ItemNav = <D,>({ info, config, additionalData,
     useDocEvent("mouseup",    () => setCurPressedObj(undefined));
     useDocEvent("mouseleave", () => setCurPressedObj(undefined));
     function handleItemNavDrag() { // Issue #478
-        if (!currentlyPressedObj || !(currentlyPressedObj instanceof Component))
+        if (!currentlyPressedObj || !(currentlyPressedObj instanceof Component)) {
+            // Disable selection while placing object - resolves #863
+            info.selections.setDisabled(true);
             return;
+        }
```
LeonMontealegre, line 90: Is this necessary if you're just deselecting everything in the SelectionPopup? I may have forgotten how our conversation went -- i remember deciding on the shared state for `isPlacing` thing but what else did we decide the exact behavior should be?

LeonMontealegre (changes requested):

src/site/shared/containers/SelectionPopup/index.tsx
riedee: I probably should have written what we talked about down - I wasn't sure if we had specifically mentioned that, so I included it because I was having trouble with rendering otherwise (the green selection circle would not disappear for some reason). I will definitely take another look with clickThrough now though, because I didn't see that when I was trying to figure out the best way to have the popup disappear.

LeonMontealegre (approved):

PR #1046, opened 2022-04-15
History box optional extra info

### LINKED ISSUE #976: History Box: Optionally displayed extra information
Some actions, like moving a component, do not display many details in the history box. Information such as what is being moved and where to/from it is being moved are stored in the action and could thus also be displayed in the history box.

Information like this can get quite detailed and clutter up the box quickly. Thus, whatever is implemented in #975 for group actions to be collapsible should also be useable to hide this information, and probably hide it by default.

jssunray34:
I added the ability for specifying custom information about `GroupAction`s and `Action`s. This additional information is presented when the user clicks on the "info" icon next to the corresponding `GroupAction` or `Action`.

Fixes #976


src/app/core/actions/GroupAction.ts
```diff
@@ -51,6 +53,12 @@ export class GroupAction implements Action {
         return `Grouped ${this.actions.length} actions` ;
     }
 
+    public getCustomInfo(): string[] | undefined {
+        if (this.customInfo)
+            return this.customInfo;
+        return undefined;
```
TGCrystal, line 59: I think this `if` is redundant, you should be able to just directly return `this.customInfo` since if it is undefined, then undefined will still be returned

TGCrystal (changes requested):

src/site/shared/containers/HistoryBox/index.tsx
```diff
@@ -17,23 +17,42 @@ type HistoryEntryProps = {
...
+                    {a.getName()}
+                </span>
+            </div>
+            {!displayExtraInfo && a.getCustomInfo &&
+                a.getCustomInfo()?.map((obj, i) => {
+                    return (
+                        <div key={`entry-extrainfo-${i}`} className="historybox__entry__extrainfo">{obj}</div>
+                    )
+                })
+            }
```
LeonMontealegre, line 45: you can condense this to:
```tsx
{!displayExtraInfo && a.getCustomInfo()?.map((obj, i) =>
    <div ..... >{obj}</div>
)}
```

src/site/shared/containers/HistoryBox/index.tsx
```diff
@@ -47,12 +66,29 @@ const GroupActionEntry = ({ g }: GroupActionEntryProps) => {
...
+                    &rsaquo;
+                </span>
+            </div>
+            {displayExtraInfo &&
+                g.getCustomInfo()?.map((obj, i) => {
+                    return (
+                        <div key={`group-action-extrainfo-${i}`} className="historybox__groupentry__extrainfo">{obj}</div>
+                    )
+                })
+            }
```
LeonMontealegre, line 91: see above

LeonMontealegre (changes requested):
lookin good! just a few formatting nits!!

src/site/shared/containers/HistoryBox/index.tsx
jssunray34: If I don't include `a.getCustomInfo` in the condition, I get the error `Cannot invoke an object which is possibly 'undefined'.`.

jssunray34: However, I am a little unsure why because I did not need this check in the below condition.

LeonMontealegre: Sorry for responding so late, for the first one, I lied -- it should've been `a.getCustomInfo?.()?.map(...)` which works correctly and the reason for the removing the `return` not working is presumably because the curly braces were still left in when they should also be deleted

LeonMontealegre (approved):

PR #1047, opened 2022-04-15
Preview objects that are Dragged and dropped in the navbar to be deleted

alstar555:
Added a preview for objects that are dragged and dropped into the navigation bar to be deleted
fixedIssueNavPreview
.


src/site/shared/containers/ItemNav/index.tsx
```diff
@@ -69,6 +71,9 @@ export const ItemNav = <D,>({ info, config, additionalData,
 
     const [hovering, setHover] = useState("");
 
+    //track whether mouse is over Nav bar
```
TGCrystal, line 74: nit: add space after `//`

src/site/shared/containers/ItemNav/index.tsx
```diff
@@ -93,6 +98,39 @@ export const ItemNav = <D,>({ info, config, additionalData,
...
+            return null;
+        if (!(currentlyPressedObj instanceof Component))
+            return null;
+        if(!overNav){
+            return null;
+        }
+        const id = GetIDFor(currentlyPressedObj);
+        var section_id;
+        var config_imgRoot;
+        var item_icon;
```
TGCrystal, line 112: nit: use `let` instead of `var`

src/site/shared/containers/ItemNav/index.tsx
```diff
@@ -93,6 +98,39 @@ export const ItemNav = <D,>({ info, config, additionalData,
         // Else just delete
         info.history.add(CreateDeleteGroupAction(info.designer, [currentlyPressedObj]).execute());
     }
+    function deletingItemPreview(){
+        if (!currentlyPressedObj)
+            return null;
+        if (!(currentlyPressedObj instanceof Component))
+            return null;
+        if(!overNav){
```
TGCrystal, line 106: nit: add space between `if` and `(`

src/site/shared/containers/ItemNav/index.tsx
```diff
@@ -183,12 +221,13 @@ export const ItemNav = <D,>({ info, config, additionalData,
...
         <div className="itemnav__preview"
              style={{
-                display: (curItemImg ? "initial" : "none"),
-                left: pos.x,
-                top: pos.y,
-             }}>
+                 display: (curItemImg ? "initial" : "none"),
+                 left: pos.x,
+                 top: pos.y,
+                }}>
```
TGCrystal, line 230: Not sure if linting said to change this spacing, but if it didn't then it should be what it was before

TGCrystal (changes requested):
Functionality looks good to me. It doesn't work on mobile, but neither does dragging to the itemnav to delete at all. Fixing that would probably be outside the scope of this pr, so you don't have to worry about it.

There is probably some way that the general logic should be refactored instead of relying on a large nested function. I'm not sure what exactly that is, so you should consult @LeonMontealegre on that.

LeonMontealegre (dismissed):

Issue #1041, opened 2022-04-01
Add a new circuit button at the left menu.


JackScofield:
If we want to create a new circuit, we need to refresh the page and then we can type the name for the new circuit. Another method is to copy the current circuit and rename it. This is not very good for using refreshing to create new circuits or copy the existing ones. 

It is good to have a "new" bottom at the left menu bar beside "My Circuits"

<img width="1505" alt="image" src="https://user-images.githubusercontent.com/37141563/161340891-881fec71-7895-4ca1-933b-0a200a5b6730.png">

PR #1049, opened 2022-04-23
Add "New Circuit" Button

EricW123:
solves #1041

### DIFF
--- a/src/site/shared/containers/SideNav/index.tsx
+++ b/src/site/shared/containers/SideNav/index.tsx
@@ -37,30 +37,14 @@ function LoadExampleCircuit(data: CircuitMetadata): Promise<string> {
 type Props = {
     helpers: CircuitInfoHelpers;
     exampleCircuits: CircuitMetadata[];
-    info: DigitalCircuitInfo;
 }
-export const SideNav = ({ helpers, exampleCircuits, info }: Props) => {
-    const { history, designer, selections } = info;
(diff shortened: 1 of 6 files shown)


src/site/shared/containers/SideNav/index.tsx
```diff
@@ -31,14 +37,31 @@ function LoadExampleCircuit(data: CircuitMetadata): Promise<string> {
...
+        if (!isSaved.isSaved && 
+            window.confirm("Current circuit not saved, do you want to save?")) {
+            helpers.SaveCircuitRemote();
+        }
+        CreateGroupSelectAction(selections, designer.getObjects()).execute();
+        const objs = selections.get().filter(s => s instanceof IOObject) as IOObject[];
+        new GroupAction([
+            CreateDeselectAllAction(selections),
+            CreateDeleteGroupAction(designer, objs)
+        ], "Cut (Context Menu)").execute()
```
TGCrystal, line 61: The group action customName parameter is optional, so the `"Cut (Context Menu)"` parameter can be deleted since this action isn't being added to the history (and the history is getting wiped immediately after anyways).

TGCrystal (changes requested):
Overall, the code looks good. There are a few functional concerns I have though.

1. The prompt to save the circuit should not appear if the user is not signed in, since the saving doesn't work anyways.
2. There should be another confirmation prompt that just confirms that the user meant to press the button so that the user does not accidentally delete what they currently have.

LeonMontealegre (approved):

PR #1050, opened 2022-04-26
Groups of items are now able to be individually pressed with a shift+click

### LINKED ISSUE #1043: Place components down individually after picking them up
When you pick up a stack of the same component, it would sometimes be easier for the user to have the option to place down the components picked up one by one instead of placing them down all at the same time.

A solution to this could be right clicking or alternate key to place one at a time instead of dropping them all at once.

Right clicking is the most intuitive option, but an alternative hot key could be used as well to place them one by one.

mattsweet417:
Since groups of items are only able to be placed together all at once, I added the functionality to place some individually by holding shift and clicking. I changed the Item Nav click function for placing items to check for a shift key press. Additionally, the reset listener was changed to only reset if shift isn't pressed.

Fixes #1043


src/site/shared/containers/ItemNav/index.tsx
```diff
@@ -15,18 +15,20 @@ import {DeleteHandler} from "core/tools/handlers/DeleteHandler";
...
 import {useWindowSize} from "shared/utils/hooks/useWindowSize";
 
-import {OpenItemNav, CloseItemNav, CloseHistoryBox, OpenHistoryBox} from "shared/state/ItemNav";
+import {OpenItemNav, CloseItemNav, CloseHistoryBox, OpenHistoryBox, itemNavReducer} from "shared/state/ItemNav";
 
 import {Draggable} from "shared/components/DragDroppable/Draggable";
 import {DragDropHandlers} from "shared/components/DragDroppable/DragDropHandlers";
 
 import "./index.scss";
+import { boolean } from "yargs";
```
TGCrystal, line 31: I'm going to guess this import was accidentally added by your editor and it can probably be safely deleted

TGCrystal (dismissed):

LeonMontealegre (approved):

PR #1071, opened 2022-06-19
yarn start now defaults to Digital

### LINKED ISSUE #1070: Make `digital` the default `yarn start` option
This should be a quite easy fix, in `scripts/start.js`

DeBestTrap:
Loops with a counter through the prompts to find Digital and breaks when found.
Fixes #1070


scripts/start.js
```diff
@@ -34,13 +34,21 @@ function start_client(dir) {
...
 // CLI
 (async () => {
+    // Get the start value for the Digital page so that "yarn start" starts on Digital
+    start_value = 0;
+    for (dir of getDirs(true, false)) {
+        if (dir.title === "Digital") {
+            break;
+        }
+        start_value += 1;
+    }
```
LeonMontealegre, line 44: Nice job! There's actually a really neat way to condense this though in JavaScript using the `findIndex` function, simply:
```js
const start_value = getDirs(true, false).findIndex((dir) => dir.title === "Digital");
```
Also note that in JS you should always declare your variables with `let` or `const` (i.e. `let start_value = 0`) otherwise it has a global scope

Also also, since we're calling `getDirs` twice, we should put at the top, a simple `const dirs = getDirs(true, false);` and reuse that variable

LeonMontealegre (changes requested):
Nice job! A few little tips though and some ways to condense it!

LeonMontealegre (approved):

Issue #1057, opened 2022-05-16
Ability to rotate a group of components individually


LeonMontealegre:
**Is your feature request related to a problem? Please describe.**
There is no way to rotate a bunch of components around themselves at the same time. Rotating a group of components rotates them around a central point together.

**Describe the solution you'd like**
Rotating a group of selections while holding a key (maybe like Shift) will rotate them individually

**Additional context**
With these selections:
![Screen Shot 2022-05-16 at 7 22 40 PM](https://user-images.githubusercontent.com/29997057/168697949-f9e7760b-8ddb-4b23-807c-99be577cb964.png)
Current behavior with rotating 90 degrees:
![Screen Shot 2022-05-16 at 7 22 47 PM](https://user-images.githubusercontent.com/29997057/168697970-fd9c7db3-ea86-4ef0-a42b-aa6fb38f24fd.png)
Target behavior with rotating 90 degrees + holding shift:
![Screen Shot 2022-05-16 at 7 23 09 PM](https://user-images.githubusercontent.com/29997057/168697996-c8c93b33-032f-4e88-9eff-0b15f9fb3e2c.png)

PR #1077, opened 2022-06-24
Multiple object independent rotation

ZWilliams752:
Added functionality to individually rotate groups of selected objects

fixes issue #1057


LeonMontealegre (changes requested):
Awesome, awesome start!

I have a few little nits and then also one major problem:

Try to do the indiviudal rotation and then undo it, and then redo it. It doesn't exactly work great.

**Before** you try and fix this, I think this would be a great time to write unit tests.

To do this, go to
`src/app/tests/digital/utils/tools/RotateTool.test.ts` and try and see if you can add some tests to test this new behavior!!

Specifically, let's have the following tests:
1. Rotation of a single object while pressing "z" (should work the same as rotation without pressing z)
2. Rotation of multiple objects (you can just do like 2 or 3) while pressing "z" and make sure to Undo and Redo it and expect the proper output
3. Rotation of a single object, first without pressing "z" and then pressing "z" without letting go of the mouse and then continuing to rotate
4. Same as 3 but for multiple objects (again make sure to Undo/Redo for all of these tests)

TGCrystal (changes requested):
Sorry, I missed your pr when I was going over them a few days ago! Looks great, just a few minor formatting nitpicks and one potential major concern.

When rotating the group as a whole, then while still rotating you press and hold `z`, it continues to rotate the group, rather than each component individually. Similarly, releasing `z` while rotating the components individually will continue to do so rather than switch to rotating them as a group. I think it would make more sense that you could press/release `z` while rotating and the rotation type would adjust mid rotation.

@LeonMontealegre's initial issue and also his comment on this pr don't explicitly say what he had in mind for this, but pressing shift for the "snap" rotation does work how I described. Pending further clarification from Leon, I would still approve this pr with the current behavior as long as the code styling nits are fixed, but it may be worth looking into changing the behavior since he may prefer it that way.

LeonMontealegre (approved):

LeonMontealegre (approved):

PR #1082, opened 2022-07-10
Frontend testing (React)

TGCrystal:
Currently adds React testing. This pr will NOT add canvas testing.


.github/workflows/ci_tests.yml
```diff
@@ -53,6 +53,36 @@ jobs:
     - name: Run tests (App) [TypeScript]
       run: yarn test  --ci app
 
+  test-digital:
```
LeonMontealegre, line 56: While this'll be faster in terms of the total-test-time, it will clog up the Checks in each PR, we should probably just separate frontend vs non-frontend tests, so `test-app` and `test-frontend` only

scripts/utils/getAliases.ts
```diff
@@ -30,7 +30,10 @@ export default function getAliases(cwd = process.cwd(), format: "webpack" | "jes
                 aliases[name] = url;
             } else {
                 const name = n.replace("/*", "/(.*)$");
-                const url = p.replace("./", "<rootDir>/").replace("/*", "/$1");
+                const url = (p.startsWith("./")
+                    ? p.replace("./", "<rootDir>/")
+                    : "<rootDir>/".concat(p)
+                ).replace("/*", "/$1");
```
LeonMontealegre, line 36: Maybe put a comment here? I forget why this was even necessary

src/site/pages/digital/tests/containers/ExprToCircuitPopup.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+// beforeAll and beforeEach can be used to avoid duplicating store/render code, but is not recommended
+//  see: https://testing-library.com/docs/user-event/intro
+describe("Main Popup", () => {
+    const info = Setup();
+    const store = createStore(reducers, applyMiddleware(thunk as ThunkMiddleware<AppState, AllActions>));
+    const user = userEvent.setup();
+
+    beforeEach(() => {
+        render(<Provider store={store}><ExprToCircuitPopup mainInfo={info} /></Provider>);
+        act(() => {store.dispatch(OpenHeaderPopup("expr_to_circuit"))});
```
LeonMontealegre, line 33: unnecessary `{` and `}` since it's a singe-statement

src/site/pages/digital/tests/containers/ExprToCircuitPopup.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+// beforeAll and beforeEach can be used to avoid duplicating store/render code, but is not recommended
+//  see: https://testing-library.com/docs/user-event/intro
+describe("Main Popup", () => {
+    const info = Setup();
+    const store = createStore(reducers, applyMiddleware(thunk as ThunkMiddleware<AppState, AllActions>));
+    const user = userEvent.setup();
+
+    beforeEach(() => {
+        render(<Provider store={store}><ExprToCircuitPopup mainInfo={info} /></Provider>);
```
LeonMontealegre, line 32: Does `render` completely reset and start with only this? or does it add to the previous `render` call? i.e. are there a bunch of these virtually being added before each test?

src/site/pages/digital/tests/containers/ExprToCircuitPopup.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+        expect(screen.getByText("Digital Expression To Circuit Generator")).not.toBeVisible();
+
+        // Reopen and requery in case reference changed
+        act(() => {store.dispatch(OpenHeaderPopup("expr_to_circuit"))});
+        expect((screen.getByPlaceholderText("!a | (B^third)") as HTMLInputElement).value).toBe("");
+    });
+
+    test("Generate Button", async () => {
+        // Enter the expression and generate
+        await user.type(screen.getByPlaceholderText("!a | (B^third)"), "a | b");
```
LeonMontealegre, line 92: that's a gross way to `get` it by? any better way?

src/site/pages/digital/tests/containers/ExprToCircuitPopup.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+    test("Cancel Button Cancels", async () => {
+        await user.type(screen.getByPlaceholderText("!a | (B^third)"), "a | b");
+
+        await user.click(screen.getByText("Cancel"));
+        expect(screen.getByText("Cancel")).not.toBeVisible();
+        expect(screen.getByText("Digital Expression To Circuit Generator")).not.toBeVisible();
+
+        // Reopen and requery in case reference changed
+        act(() => {store.dispatch(OpenHeaderPopup("expr_to_circuit"))});
```
LeonMontealegre, line 86: unnecessary { and } since it's a singe-statement

src/site/pages/digital/tests/containers/ExprToCircuitPopup.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+        expect(led).toBeDefined();
+        expect(led.isOn()).toBeFalsy();
+        inputA.click();
+        expect(led.isOn()).toBeTruthy();
+        inputA.click();
+        inputB.click();
+        expect(led.isOn()).toBeTruthy();
+
+        // Reopen and requery in case reference changed
+        act(() => {store.dispatch(OpenHeaderPopup("expr_to_circuit"))});
```
LeonMontealegre, line 118: unnecessary { and } since it's a singe-statement

src/site/pages/digital/tests/containers/ExprToCircuitPopup.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+        expect((screen.getByPlaceholderText("!a | (B^third)") as HTMLInputElement).value).toBe("");
```
LeonMontealegre, line 119: once again, is there a better way to query it? if not, better to maybe make a constant at the top since it's such a specific string?
also preferrable (if possible):
`screen.getByPlaceholderText<HTMLInputElement>("!a | (B^third)")`

src/site/shared/tests/helpers/GetToggles.ts
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * @param id The id of the text to search for.
+ * @returns    An array containing the pressed and unpressed images, respectively.
+ */
+export function GetToggles(id: Matcher): [HTMLImageElement, HTMLImageElement] {
+    const buttons = screen.getByText(id).parentNode!.querySelectorAll("img");
+    return buttons[0].src.includes("Down")
+        ? [buttons[0], buttons[1]]
+        : [buttons[1], buttons[0]];
+}
```
LeonMontealegre, line 15: You are pretty much strictly using this helper to then check to make sure that the Toggle is on/off, you should probably instead just have a helper that gets the state of the toggle, and then `expect` that state to be on/off. Then in the actual ButtonToggle/SwitchToggle tests, that's where the functionality of the images being correspondingly on/off should be tested

src/site/shared/tests/components/ButtonToggle.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+    test("Button off", () => {
+        render(<ButtonToggle text="test" isOn={false} />);
+        const [buttonOn, buttonOff] = GetToggles("test");
+        expect(buttonOn).not.toBeVisible();
+        expect(buttonOff).toBeVisible();
+    });
+    test("onChange", async () => {
+        let isOn = false;
+        const user = userEvent.setup();
+        render(<ButtonToggle text="test" isOn={isOn} onChange={() => isOn = !isOn} />);
```
LeonMontealegre, line 26: This test implies that you're changing whether or not the Button is on/off, but setting `isOn` doesn't actually change it to be on/off cause it's not a state variable, so make a different name for the variable

LeonMontealegre (changes requested):

scripts/utils/getAliases.ts
TGCrystal: Tried undoing it and I guess it wasn't even necessary

src/site/pages/digital/tests/containers/ExprToCircuitPopup.test.tsx
TGCrystal: Removing them gives the following error:
```
Overload 1 of 2, '(callback: () => Promise<void>): Promise<undefined>', gave the following error.
  Type '{ readonly type: "OPEN_HEADER_POPUP_ID"; readonly popup: HeaderPopups; }' is missing the following properties from type 'Promise<void>': then, catch, finally, [Symbol.toStringTag]        
Overload 2 of 2, '(callback: () => VoidOrUndefinedOnly): void', gave the following error.
  Type '{ readonly type: "OPEN_HEADER_POPUP_ID"; readonly popup: HeaderPopups; }' is not assignable to type 'VoidOrUndefinedOnly'.
```

src/site/pages/digital/tests/containers/ExprToCircuitPopup.test.tsx
TGCrystal: The library does some automatic cleanup between tests and basically derenders everything, so it is necessary to do this before each test

src/site/pages/digital/tests/containers/ExprToCircuitPopup.test.tsx
TGCrystal: Turns out there was a better way, I didn't think getByRole would work since it isn't a div but it did.

src/site/shared/tests/helpers/GetToggles.ts
TGCrystal: What would I use to check if it on/off? I would think it would be checking if the "down" button is visible or not, but the `toBeVisible` thing I've been using to test checks a few different conditions: https://github.com/testing-library/jest-dom#tobevisible

src/site/pages/digital/tests/containers/ExprToCircuitPopup.test.tsx
LeonMontealegre: Looks like an open issue https://github.com/microsoft/TypeScript/issues/12871, just add spaces around braces at least then

src/site/pages/digital/tests/containers/ExprToCircuitPopup.test.tsx
LeonMontealegre: Okay cool

src/site/shared/tests/helpers/GetToggles.ts
LeonMontealegre: If the expected image is visible or not yeah

LeonMontealegre (approved):

PR #1090, opened 2022-07-22
Replace component

### LINKED ISSUE #502: "Replace Component" Button
It can be annoying if you've wired a component up and want to replace it with a different component, say you used an ANDGate and wanted an ORGate, but to replace it you'd have to delete it (which removes all existing connections) and then place the ORGate and reconnect everything

There should be a button on the Selection Popup called "Replace" which brings up a menu showing all compatible components to replace with.

Ex. If I wanted to replace an ANDGate with 2 input ports (1 output port), it would find each component that can have 2 input ports and 1 output port.
Similarly, if you wanted to replace a Switch (0 input ports, 1 output port), it could replace it with something like a BUFGate or even an AND or OR gate since they each have 1 output port.

So more formally, compatible components would be ones with more or equal amounts of ports, ordered by how close to the number of ports it is.

TGCrystal:
Closes #502


LeonMontealegre (changes requested):
Bugs:

1. There isn't spacing between the "Replace Component" text and the "View IC" button, probably other buttons too:
Screen Shot 2022-08-31 at 3 03 17 PM



2. Replacing a 3-Selector Mux with an Encoder causes funkiness:
Screen Shot 2022-08-31 at 3 06 41 PM

To reproduce:
- Place a Mux
- Increase select port count to 3
- Replace it with an Encoder

I believe the Encoder shouldn't even be able to replace the Mux since an encoder with 1 output has 2 inputs. This might be a difficult bug to fix

A similar issue is if you replace a D-latch with an Encoder, the encoder has 2 outputs but the input amount is just 2 when it should be 4... which isn't compatible with a D-latch anyways. Definitely something funky happening here.

LeonMontealegre (changes requested):
just need to fix test cases and we should be all good

LeonMontealegre (approved):

LeonMontealegre (approved):

PR #1092, opened 2022-07-23
Able to bus components

### LINKED ISSUE #1056: Add bus option when selecting components
It can be frustrating to have to select specifically Ports for the purpose of Bus'ing them together.

Commonly, the user would want to simply bus all the components they have selected together, so it would be nice that if the user had components selected with equal number of input/output ports, it would bus them.

**Additional context**
There would be ambiguity in the case of two components with both the same number of inputs and outputs, so it should probably be restricted such that one of the selections needs only inputs/outputs or potentially if they are both equal, it won't show the option since it is ambiguous. 

**Examples**
Where bus'ing would be allowed:
- Constant Number and a BCD Display
- Four Switch's and a BCD Display
- Four Switch's and two 2-input ANDGates
- Six Switch's and a Mux

Where it would *not* be allowed:
- Two BUFGates
- Three Switch's and a BCD Display
- Seven Switch's and a Mux

DeBestTrap:
Fixes #1056

Splits components into three categories: only inputs, only outputs, ambiguous. Can bus components that are in two of these categories but not all three.

Also retains the previous selecting individual ports to bus but will only work if no components are selected.


DeBestTrap:
Fixed more issues that cropped up due to some testing I did.

src/site/pages/digital/src/containers/SelectionPopup/modules/BusButtonModule.tsx
```diff
@@ -13,23 +16,83 @@ type Props = {
...
-    const [props, ps] = useSelectionProps(
+    const [props, items] = useSelectionProps(
         info,
-        (s): s is (InputPort | OutputPort) => (
-            (s instanceof InputPort && s.getWires().length === 0) ||
-            s instanceof OutputPort
+        (s): s is (Component | InputPort | OutputPort) => (
+            (s instanceof Component ||
+             s instanceof InputPort && s.getWires().length === 0) ||
+             s instanceof OutputPort && s.getWires().length === 0
```
TGCrystal, line 24: The `&&` operator executes before the `||`, so this here works fine as is. For clarity, I think there should be parenthesis around each of the `&&` operands. Also, the parenthesis there does not have to go around `s instanceof Component`.
Basically, this block should look like:
```ts
s instanceof Component ||
(s instanceof InputPort && s.getWires().length === 0) ||
(s instanceof OutputPort && s.getWires().length === 0)
```

src/site/pages/digital/src/containers/SelectionPopup/modules/BusButtonModule.tsx
```diff
@@ -13,23 +16,83 @@ type Props = {
...
+        const ambiComps = components.filter(s => s.getPorts().some(p => p instanceof InputPort &&
+                                                                        p.getWires().length === 0) ===
+                                                 s.getPorts().some(p => p instanceof OutputPort &&
+                                                                        p.getWires().length === 0)) as Component[];
+        // Remove ambiguous components that have no available ports
+        //  (occurs when an input/output only component is already connected to other components
+        //   and thus 0 free iports === 0 free oports)
+        ambiComps.forEach(c => {
+            if (c.getPorts().filter(p => p instanceof InputPort && p.getWires().length === 0).length === 0 &&
+                c.getPorts().filter(p => p instanceof OutputPort && p.getWires().length === 0).length === 0) {
```
TGCrystal, line 54: I don't think there should be a need to separately check that all the `InputPort`s and `OutputPort`s are full, since `.getPorts()` is returning type `Port` and the only subclasses it has in digital are `InputPort` and `OutputPort`. So I think the condition can be simplified to `c.getPorts().filter(p => p.getWires().length === 0).length === 0` and then a bit further to `c.getPorts().every(p => p.getWires().length > 0)`

TGCrystal (changes requested):

DeBestTrap:
Applied the requested changes

TGCrystal (dismissed):

LeonMontealegre (approved):

TGCrystal (approved):

PR #1113, opened 2022-09-15
Unit Normalization and Property Condensing

### LINKED ISSUE #751: Circuit Position Y-Axis Flip
The y-axis for the position of circuits has down being positive. However, to non-cs folks, hitting the up arrow to increase the y-value of the position and having the circuit move down will be counterintuitive. Pressing up for position should move the circuit up, so up should be positive.

When implementing this change, you might need to have some kind of flag for file saves for before & after this change because existing projects would have their circuit locations mirrored over the x-axis if all y values were suddenly negated.

LeonMontealegre:
Renormalized all the units that OpenCircuits uses such that a grid cell is now `1` unit instead of `50`, and makes positioning information more intuitive.

Also flipped the y-axis such that +y is up which will be more intuitive and fixes #751 

Finally, condensed a lot of properties and removed selection popup modules that are no longer necessary

Note that this is currently a **breaking** change since I didn't finish writing the migrators fully.
This will be done in the next installment of the model refactor PRs.

To migrate:
- SegmentDisplay, `segmentCount` to Props system
- Clock, `isOn` to Props system
- Switch + Button, `on`, `pressableBox`
- IOObject, `name` to props system
- FlipFlop, `state`, `clock`, `lastClock` to Props system
- Latch, `enabled`, `state` to Props system
- Gate, `not` to Props system
- Unit scaling (1/50, 1/50)
- Y-axis flip


TGCrystal (changes requested):
Looks mostly good. Only issue I see is that the flipped y direction is not accounted for by smart place or the expression to circuit generator. Should be an easy fix for both.

TGCrystal (approved):

PR #1115, opened 2022-09-18
Lint only changed files

TGCrystal:
This is branched from #1114


scripts/lint.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    const files: string[] = [];
+    if (!all) {
+        const git = simpleGit();
+        const diff = await git.diff(["--name-only", "origin/master...", "--"]);
+        diff.split("\n").filter((file) => (file.endsWith(".ts") || file.endsWith(".tsx"))).forEach((file) => {
+            if (file.endsWith(".ts") || file.endsWith(".tsx"))
+                files.push(file);
+        });
+        const status = await git.status(["--short"]);
```
LeonMontealegre, line 28: Why do you need to do this as well?

LeonMontealegre (approved):

PR #1116, opened 2022-09-18
Use a cache for `yarn` in GitHub Actions

TGCrystal:


TGCrystal:
On further inspection, this isn't caching `node-modules` and we already commit `yarn.lock` so I don't think it is actually doing anything. See: https://github.com/actions/setup-node#caching-global-packages-data

PR #1185, opened 2022-10-18
Constant High

Rcheng2003:
Added the component Constant High( Issue #1127)


TGCrystal (changes requested):
The initial code here looks good, but the component itself does not look correct and is missing the correct behavior once placed. See http://www.opencircuits.io/ to see how it currently functions

PR #1186, opened 2022-10-21
Add back buf gate

ragar3:


TGCrystal (changes requested):
Looks good, only thing to change is making the line connecting the node to the gate connect better. 
image
Here is what it looks like on master for comparison:
image

ragar3:
Resolved change request + merge conflicts

TGCrystal (dismissed):
Looking good on that side!

Sorry to be this pedantic, but there is a minor visible seam between the line on the left side and the gate itself. Should just need a change like the one you did on the other side, but very minor this time.
image

src/app/core/views/info/digital.ts
```diff
@@ -0,0 +1,25 @@
+import {ANDGate, DigitalComponent, DigitalNode,
```
LeonMontealegre, line 1: This file should be deleted

src/app/digital/views/components/BUFGateView.ts
```diff
@@ -0,0 +1,16 @@
...
+import {ComponentView} from "core/views/ComponentView";
+
+import {DigitalViewInfo} from "../DigitalViewInfo";
+
+
+export class BUFGateView extends ComponentView<BUFGate, DigitalViewInfo> {
+    public constructor(circuit: DigitalViewInfo, obj: BUFGate) {
+        super(circuit, obj, V(1, 1), "buf.svg");
+    }
+}
```
LeonMontealegre, line 16: You don't need a custom view here, you can just do `new ComponentView<BUFGate>(c, o, V(1, 1), "buf.svg")` in `src/app/digital/views/index.ts`
(since all this class does is call the `super` constructor which is `ComponentView` and has no extended functionality)

LeonMontealegre (changes requested):
Looking good! Just a few things that can be simplified and a file that's no longer necessary

LeonMontealegre (changes requested):
One final small change and we should be good to merge!

LeonMontealegre (approved):

PR #1188, opened 2022-10-24
Fixed backwards propagation

### LINKED ISSUE #1181: Model Refactor: Backwards Connection Propagation is Weird
**Describe the bug**
Splitting an 'On' wire that was connected from input to output causes half the wire to oscillate on/off rapidly and the other to turn off

**To Reproduce**
1. Go to model refactor site
2. Add switch
3. Add ANDGate
4. Turn switch on
5. Connect from ANDGate input to switch output
6. Split wire
7. Observe the oscillation

**Expected behavior**
Both sections of the split wire should remain in a stable on state.

**Screenshots**
backwardsProp

 - OS: Windows
 - Browser: Chrome
 - Version: 106.0.5249.91

*PLEASE MAKE SURE TO APPROPRIATELY LABEL YOUR ISSUE, YOUR ISSUE IS REQUIRED TO SPECIFY AT LEAST A GENERAL IDEA OF: Area, Difficulty, Priority, and Scope! If the issue is specific to digital/analog, please also specify that with the appropriate label!

tingll99:
Added a simple check for if `p2` is an output wire, which then swaps the order on the createWire call. It doesn't look at input/select so that distinction should not cause issues in the future

Closes #1181


TGCrystal (dismissed):

src/app/core/tools/WiringTool.ts
```diff
@@ -91,8 +93,14 @@ export const WiringTool = (() => {
 
             const port2 = findPort(info);
             if (port2 !== undefined) {
-                const wire = CreateWire(circuit.getWireKind(), port!.id, port2.id);
-                history.add(Place(circuit, wire));
+                if (port2.group === DigitalPortGroup.Output) {
```
LeonMontealegre, line 96: Gotta get back up-to-date with `model_refactor` since groups are now strings
Also add a TODO at the top saying this shouldn't be necessary in the future hoepfully

src/app/core/tools/WiringTool.ts
```diff
@@ -90,9 +90,21 @@ export const WiringTool = (() => {
             const { circuit, history } = info;
 
             const port2 = findPort(info);
+
+
+            //TODO
+            //the below group check shouldn't be necessary in future versions
+            //currently CreateWire doesn't determine input vs output
+            //so we do a quick check here. CreateWire should eventually though
```
LeonMontealegre, line 98: nit: put spaces after each `//`

src/app/core/models/types/base/Port.ts
```diff
@@ -7,7 +7,7 @@ export type Port = BaseObject & {
     baseKind: "Port";
 
     parent: GUID;
-    group: string; // The group for the port would be like input/output/select port
+    group: string; // groups: "inputs", outputs", or "selects"
```
LeonMontealegre, line 10: not strictly true in the future, will be arbitrary

LeonMontealegre (changes requested):
just a few small nits! looking good!

PR #1189, opened 2022-10-24
Constant low #1126

rachelt323:


TGCrystal (changes requested):
Code looks good, but the Constant Low itself currently looks like
image
when it should look like
image

PR #1190, opened 2022-10-25
Add back XOR and XNOR Gates

### LINKED ISSUE #1141: Model Refactor: Add back XNORGate

mattsweet417:
Fixes #1141 and #1140


TGCrystal (approved):
Looks good to me! I'm not sure about using a dropdown instead of the number input for the number of ports, but it seems like both are broken on model_refactor anyways so I'll defer to @LeonMontealegre on this

src/app/core/views/portinfo/digital/index.ts
```diff
@@ -45,4 +45,26 @@ export const DigitalPortInfo: PortInfoRecord<DigitalComponent> = {
...
+    },
+    "XNORGate": {
+        ...DefaultDigitalPortInfo,
+        AllowChanges: true,
+        ChangeGroup:  "inputs",
+
+        // Generate configs for 2->8 input ports
+        PositionConfigs: [2,3,4,5,6,7,8].map((numInputs) => ({
+            "inputs":  CalcQuadCurvePortPositions(numInputs, 1.2, .48, V(-1, 0)),
+            "outputs": [CalcPortPos(V(.8, 0), V(.7, 0))], // 1 output
```
LeonMontealegre, line 67: I think the `outputs` here can be the same as `XORGate`, the circle for the `not` can simply lay on-top of the port
Screen Shot 2022-11-08 at 1 51 01 PM
(this is a screenshot from master)

src/app/digital/views/components/XORGateView.ts
```diff
@@ -0,0 +1,78 @@
+import {DEFAULT_BORDER_COLOR, DEFAULT_BORDER_WIDTH, DEFAULT_CURVE_BORDER_WIDTH, DEFAULT_FILL_COLOR, GATE_NOT_CIRCLE_RADIUS, SELECTED_BORDER_COLOR, SELECTED_FILL_COLOR} from "core/utils/Constants";
```
LeonMontealegre, line 1: I think you can see that these classes are extraordinarily similar to each-other.
I think a good compromise would be to have just a `XORGateView` and it can take in an extra parameter in the constructor for `not`, which when `not` is true, it will draw the not circle

LeonMontealegre (changes requested):
Looking good! Just a few changes

mattsweet417:
I added the overridden function from #1192 to draw the not circle in front of gate when rendered, so it may cause issues when merging.

mattsweet417:
But everything should be fixed up :)

LeonMontealegre (approved):

Issue #505, opened 2020-02-07
Add Circuit Sharing / Permissions


KevinMackenzie:
Currently, circuits can only be accessed by their owners.  We want circuits to be shareable with other users and publicly.   The idea is that there are two sharing systems: Share with User; Share via Link.

Sharing status with any user can be revoked or amended at any time by the owner only.  The link used to access a circuit via "Share with User" is the same as the link used by the owner.  It contains the circuit ID of the circuit.

Sharing via Link is done with a single link that can be revoked or regenerated at any time and permissions on it can be changed at any time.  The link will be just like a normal link, but the circuit ID will be a "fake" one that will redirect to a real circuit in all permitted operations.  The implementation details for this are TBD and may be tricky.

URLs for a circuit do not exist yet, so that must be added before this can be finished (see #388)

Here are the required new database features:

New table:
User Sharing:
- Entry: For each circuit: for each user shared with
- Column: CircuitID (INDEX)
- Column: UserID (INDEX)
- Column: Permission enum (view|edit)

New table:
Link Sharing:
- Entry: 0 or 1 entry per circuit
- Column: CircuitID (INDEX)
- Column: WeakCircuitID / shareID (INDEX) (used in URL)
- Column: Permission enum (view|edit)

WeakCircuitID: points to real circuit; accessible anonymously; "unguessable"; can be regenerated

PR #1191, opened 2022-10-25
Major back-end rework

### LINKED ISSUE #504: Merge the "Create" and "Update" paths for circuits

### LINKED ISSUE #710: Switch to env variables instead of flags for backend

### LINKED ISSUE #1157: Add support for multi-part circuits in GCP Datastore
[GCP Firestore in Datastore mode](https://cloud.google.com/datastore/docs/), our production database configuration, does not support entities (database rows, circuits in our case) larger than 1MB (See documentation on [limits](https://cloud.google.com/datastore/docs/concepts/limits)).  We anticipate circuits that are larger than 1MB, so we need to add support for splitting a single circuit across multiple entities.

Suggested model is:
Each root circuit entity in the database can contain a list of child circuit ID's.  If this list is non-empty, the back-end should load said list and concatenate its content with the root circuit's content.  This construct should be completely isolated to the  GCP storage back-end.

KevinMackenzie:
- Merge insert/update paths (Fixes #504)
- Parameters to server can be specified as env vars or CLI flags (Fixes #710)
- Add key-based circuit sharing (circuit ID + secret key = access) (Preliminary work for #505)
- Removes implicit 1MB circuit limit in datastore (up to ~30MB) (Fixes #1157)
- Removes `sqlite`
- Adds tests for GCP datastore / mem database interfaces
- Adds e2e API tests

Currently targeting master with these changes, but that may change.


src/server/db/datastore/v0_1/types.go
```diff
@@ -0,0 +1,74 @@
...
+	ContentID ContentID
+	Content   string `datastore:",noindex"`
+}
+
+func NewContentHeaderKey(circuitKey *datastore.Key) *datastore.Key {
+	return datastore.NameKey(CircuitContentHeaderKind, "Content", circuitKey)
+}
+
+// ContentHeader is the authoratative source for which seqeuence of ContentChunk's is current
+type ContentHeader struct {
```
KevinMackenzie, line 59: A thought here: If we kept multiple of these objects in the DB, then we can create a coarse history of the circuit.  Not hard to add later on (not a breaking change), but I have to think more if this is sufficient for storing "milestone circuits".

KevinMackenzie: Thought about this some more.  We still have only one `ContentHeader` pointing to the "current" version.  We do need 
some kind of timestamp needs to be added to the first `ContentChunk` entry of a given `ContentID` to establish a timeline once `ContentHeader` no longer points to it.

PR #1192, opened 2022-10-25
Model refactor nand

aruan20:


TGCrystal (commented):
The logic doesn't seem to work when using more than one import port, but it is hard to consistently test since the selection popup is kind of broken (not your fault). Otherwise, looks good to me so far though

src/app/digital/views/DigitalPortView.ts
```diff
@@ -25,4 +37,26 @@ export class DigitalPortView extends PortView<DigitalPort, DigitalViewInfo> {
...
+        const { origin, target } = this.pos.get();
+
+        const lineCol       = (parentSelected && !selected ? SELECTED_BORDER_COLOR : DEFAULT_BORDER_COLOR);
+        const borderCol     = (parentSelected ||  selected ? SELECTED_BORDER_COLOR : DEFAULT_BORDER_COLOR);
+        const circleFillCol = (parentSelected ||  selected ? SELECTED_FILL_COLOR   : DEFAULT_FILL_COLOR);
+        const lineStyle   = new Style(undefined, lineCol, IO_PORT_LINE_WIDTH);
+        const circleStyle = new Style(circleFillCol, borderCol, IO_PORT_BORDER_WIDTH);
+
+        renderer.draw(new Line(origin, target), lineStyle);
+        renderer.draw(new Circle(target, IO_PORT_RADIUS), circleStyle);
```
LeonMontealegre, line 54: use `super.renderInternal` for all of this

src/app/digital/views/DigitalPortView.ts
```diff
@@ -25,4 +37,26 @@ export class DigitalPortView extends PortView<DigitalPort, DigitalViewInfo> {
...
+
+        if (this.circuit.getPortParent(this.obj).kind === "NANDGate" && this.obj.group === "outputs"){
```
LeonMontealegre, line 56: make a constant right above called `parent = this.circuit.getPortParent(this.obj)` to simplify this a bit
Also add a comment explaining why this is necessary for now

src/app/digital/views/components/NANDGateView.ts
```diff
@@ -0,0 +1,55 @@
...
+
+import {NANDGate} from "core/models/types/digital";
+
+import {RenderInfo}    from "core/views/BaseView";
+import {ComponentView} from "core/views/ComponentView";
+
+import {DigitalViewInfo} from "../DigitalViewInfo";
+
+
+export class NANDGateView extends ComponentView<NANDGate, DigitalViewInfo> {
```
LeonMontealegre, line 19: Notice how similar this is to `ANDGateView`, we probably want to re-use `ANDGateView`, so to do this, we can have `ANDGateView` extend `ComponentView<ANDGate | NANDGate` and take in an `ANDGate | NANDGate` and then re-use this in `src/app/digital/views/index.ts`

LeonMontealegre (changes requested):
Looking good! Just a few things to fix

LeonMontealegre (approved):

PR #1200, opened 2022-11-02
ORGate and NORGate added back

Kagiri2:


TGCrystal (changes requested):
2 issues I noticed:
1) The gates themselves are using the logic for AND gates rather than xor/xnor
2) The XNOR has the port behind the line. Whoever implemented the NANDGate in #1192 ran into this, Leon's solution was: "I think the solution here might be to draw the circle in DigitalPortView, so the port will draw its circle (override renderInternal, see PortView).
For it to know if it should draw the circle or not, you can get the parent component from the Port and just check if it's kind is a NANDGate and that the Port itself is an Output port and we'll eventually make a better system; but for now this is fine"

src/app/core/views/portinfo/digital/index.ts
```diff
@@ -4,7 +4,7 @@ import {V} from "Vector";
 
 import {DefaultDigitalPort, DigitalComponent} from "core/models/types/digital";
 
-import {CalcPortPos, CalcPortPositions} from "../positioning/utils";
+import {CalcPortPos, CalcPortPositions,CalcQuadCurvePortPositions} from "../positioning/utils";
```
LeonMontealegre, line 7: nit: put space after `,`

src/app/core/views/portinfo/digital/index.ts
```diff
@@ -45,4 +45,26 @@ export const DigitalPortInfo: PortInfoRecord<DigitalComponent> = {
...
+    },
+    "NORGate": {
+        ...DefaultDigitalPortInfo,
+        AllowChanges: true,
+        ChangeGroup:  "inputs",
+
+        // Generate configs for 2->8 input ports
+        PositionConfigs: [2,3,4,5,6,7,8].map((numInputs) => ({
+            "inputs":  CalcQuadCurvePortPositions(numInputs, 1.2, .48, V(-1, 0)),
+            "outputs": [CalcPortPos(V(0.8, 0), V(.7, 0))], // 1 output
```
LeonMontealegre, line 67: I think the outputs here can be the same as `ORGate`, the circle for the not can simply lay on-top of the port
Screen Shot 2022-11-08 at 2 36 31 PM
(this is a screenshot from master)

src/app/digital/views/components/NORGateView.ts
```diff
@@ -0,0 +1,88 @@
...
+import {NORGate} from "core/models/types/digital";
+import {RenderInfo}               from "core/views/BaseView";
+import {ComponentView}            from "core/views/ComponentView";
+import {Circle} from "core/utils/rendering/shapes/Circle";
+import {Renderer} from "core/utils/rendering/Renderer";
+import {QuadCurve} from "core/utils/rendering/shapes/QuadCurve";
+import {DigitalViewInfo} from "../DigitalViewInfo";
+
+
+// // sheesh
```
LeonMontealegre, line 18: ??

src/app/digital/views/index.ts
```diff
@@ -5,11 +5,13 @@ import {ViewFactory, ViewRecord}  from "core/views/ViewManager";
 import {DigitalCircuitController} from "digital/controllers/DigitalCircuitController";
 
 import {ANDGateView}     from "./components/ANDGateView";
+import {ORGateView}     from "./components/ORGateView";
 import {LEDView}         from "./components/LEDView";
 import {SwitchView}      from "./components/SwitchView";
 import {DigitalPortView} from "./DigitalPortView";
 import {DigitalViewInfo} from "./DigitalViewInfo";
 import {DigitalWireView} from "./DigitalWireView";
+import { NORGateView } from "./components/NORGateView";
```
LeonMontealegre, line 14: remove spaces and sort imports

src/app/digital/views/info/index.ts
```diff
@@ -13,4 +13,6 @@ export const DigitalPropInfo: PropInfoRecord<DigitalObj> = {
     "Switch":  DefaultComponentPropInfo,
     "LED":     DefaultComponentPropInfo,
     "ANDGate": DefaultComponentPropInfo,
+    "ORGate": DefaultComponentPropInfo,
```
LeonMontealegre, line 16: vertically align

LeonMontealegre (changes requested):
Looking good! Just a few things to do

PR #1214, opened 2022-12-26
Add eslint-tsdoc

TGCrystal:
After vacation I'll try replacing the current jsdoc generation script with something powered by the tsdoc plugin


TGCrystal:
I found an issue this had with the doc generation and it will probably have bad merge conflicts after the api refactor, so I think it will be easier to close this and redo it later

PR #1215, opened 2023-01-27
More flesh on CircuitInternal skeleton

KevinMackenzie:
Filled out parts of `CircuitInternal` to facilitate history, undo/redo, transactions.  Added a couple public functions to provide a front-end to this "CircuitOp" model for mutating the circuit.

Added sample `CircuitInfo` type so `CircuitInternal` is generic across digital/analog and doesn't need knowledge of specific components.

Future work:
- Decide on an interface for providing events/callbacks from this mutation system.

### DIFF
--- a/src/app/core/internal/impl/CircuitInternal.ts
+++ b/src/app/core/internal/impl/CircuitInternal.ts
@@ -1,72 +1,172 @@
-import {GUID} from "core/schema/GUID";
+import { GUID } from "core/schema/GUID";
 
-import {Schema}       from "../../schema";
-import {GetDebugInfo} from "../utils/Debug";
+import { Schema } from "../../schema";
+import { GetDebugInfo } from "../utils/Debug";
+import { CircuitOp, InvertMultiOp, MultiOp } from "./CircuitOps";
+import { ComponentInfoProvider, PortConfig } from "./ComponentInfo";
 
-import {DebugOptions}      from "./DebugOptions";
-import {HistoryManager}    from "./HistoryManager";
-import {SelectionsManager} from "./SelectionsManager";
-
+import { HistoryManager } from "./HistoryManager";
 
 export class CircuitInternal {
-    protected readonly circuit: Schema.Circuit;
-
-    protected readonly selections: SelectionsManager;
     protected readonly history: HistoryManager;
-
-    public debugOptions: DebugOptions;
-    public isLocked: boolean;
+    protected readonly componentTypes: ComponentInfoProvider;
 
     protected objMap: Map<GUID, Schema.Obj>;
     protected componentPortsMap: Map<GUID, Schema.Port[]>; // components to ports
     protected connectionsMap: Map<GUID, Schema.Wire[]>; // Ports to wires
 
-    public constructor() {
-        this.circuit = Schema.DefaultCircuit();
-        this.selections = new SelectionsManager();
+    // Explicit variable only needed for asserts.  JS threading model means no mutex needed, but it
+    // is possible for an async call mid-transaction to yield control-flow to another task and break
+    // atomicity of transactions.  Any behavior that yields control-flow mid transaction is ILLEGAL,
+    // and thus guarded by asserts.
+    protected transaction: boolean;
+
+    // TODO: Can the emitted events be completely generated from this Ops list?
+    protected transactionOps: CircuitOp[];
+
+    public constructor(componentsTypes: ComponentInfoProvider) {
         this.history = new HistoryManager();
+        this.componentTypes = componentsTypes;
 
         this.objMap = new Map();
         this.componentPortsMap = new Map();
+        this.connectionsMap = new Map();
 
-        this.debugOptions = {
-            debugCullboxes:       false,
-            debugNoFill:          false,
-            debugPressableBounds: false,
-            debugSelectionBounds: false,
-        };
-        this.isLocked = false;
+        this.transaction = false;
+        this.transactionOps = [];
     }
 
-    public addObj(obj: Schema.Obj) {
-        if (this.objMap.has(obj.id))
-            throw new Error(`CircuitInternal: Attempted to add obj ${GetDebugInfo(obj)} which already existed in the circuit!`);
-
-        this.circuit.objects.push(obj);
-        this.objMap.set(obj.id, obj);
+    private assertOwnObject(obj: Schema.Obj): void {
+        if (this.objMap.get(obj.id) !== obj)
+            throw new Error("Circuit does not own provided object");
     }
-    public removeObj(obj: Schema.Obj) {
-        if (!this.objMap.has(obj.id))
-            throw new Error(`CircuitInternal: Attempted to remove obj ${GetDebugInfo(obj)} which does not exist in the circuit!`);
-
-        // TODO: make this more efficient, somehow keep track of the indices of each object
-        this.circuit.objects.splice(this.circuit.objects.indexOf(obj), 1);
-        this.objMap.delete(obj.id);
+    private assertHasGUID(id: GUID): void {
+        if (!this.objMap.has(id))
+            throw new Error("Circuit does not own provided GUID");
+    }
+    // Public for use by i.e. the Renderer, Propagators.
+    public assertTransactionState(st: boolean): void {
+        if (this.transaction != st)
+            throw new Error("Unexpectedly in transaction state");
     }
 
-    public setPropFor<O extends Schema.Obj, K extends keyof O["props"]>(obj: O, key: K, val: O["props"][K]) {
+    public beginTransaction(): void {
+        this.assertTransactionState(false);
+        this.transaction = true;
+    }
+    public commitTransaction(): MultiOp | undefined {
+        this.assertTransactionState(true);
+        let op: MultiOp | undefined = undefined;
+        if (this.transactionOps.length > 0) {
+            op = { kind: "MultiOp", ops: this.transactionOps };
+            this.history.push(op);
+            this.transactionOps = [];
+
+            // TODO: Emit events for changes...
+        }
+        this.transaction = false;
+        return op;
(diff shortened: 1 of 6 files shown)


src/app/core/internal/impl/ComponentInfo.ts
```diff
@@ -0,0 +1,112 @@
...
+export function IsValidPortList(info: ComponentInfo, ports: Schema.Port[]): boolean {
+    let countMap = new Map(info.portGroups.map(g => [g, Number(0)]));
+    for (const p of ports) {
+        let count = countMap.get(p.group);
+        if (count === undefined)
+            return false;
+        count++;
+    }
+    let counts: number[] = info.portGroups.map(g => countMap.get(g) ?? -1);
+    return info.isValidPortConfig({ counts: counts });
```
LeonMontealegre, line 37: Mmmmm I don't like that `counts` needs to be an array that's 1:1 with the groups defined in `ComponentInfo`, can we not just pass this map of group types to counts in `info.isValidPortConfig`? see below

src/app/core/internal/impl/ComponentInfo.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    public get kind(): string {
+        return this._kind;
+    }
+    public readonly portGroups: string[] = ["IN", "OUT"];
+    public readonly defaultPortConfig: PortConfig = ({ counts: [2, 1] });
+
+    public isValidPortConfig(p: PortConfig): boolean {
+        // At least 2 inputs, exactly 1 output
+        return p.counts[0] >= 2 && p.counts[1] == 1;
```
LeonMontealegre, line 60: I feel like a PortConfig counts should be a `Record<string, number>` so that it can instead be written as
`p.counts["IN"] >= 2 && p.counts["OUT"] == 1`, which seems a lot more readable

src/app/core/internal/impl/HistoryManager.ts
```diff
@@ -1,5 +1,62 @@
+import { CircuitOp, InvertMultiOp, MultiOp } from "./CircuitOps";
```
LeonMontealegre, line 1: nit: import spaces

src/app/core/internal/impl/HistoryManager.ts
```diff
@@ -1,5 +1,62 @@
...
+    public push(op: CircuitOp | MultiOp): void {
+        if (this.redoStack.length > 0)
+            this.redoStack.clear();
+        this.undoStack.push(op);
+    }
+
+    // Returns a MultiOp that undoes the most recent MultiOp
+    public undo(): MultiOp | undefined {
+        let op = this.undoStack.pop();
+        if (op !== undefined) {
```
LeonMontealegre, line 45: Prefer:
```ts
if (!op)
    return undefined;
// continue with logic if op is defined
```

src/app/core/internal/impl/HistoryManager.ts
```diff
@@ -1,5 +1,62 @@
...
+            this.redoStack.push(op);
+            return InvertMultiOp(op);
+        }
+        return undefined;
+    }
+
+    // Returns a MultiOp that re-does the most recently undone MultiOp
+    public redo(): MultiOp | undefined {
+        let op = this.redoStack.pop();
+        if (op !== undefined) {
```
LeonMontealegre, line 55: same as above

src/app/core/public/api/impl/Component.ts
```diff
@@ -7,19 +6,17 @@ import {Component}     from "../Component";
 import {ComponentInfo} from "../ComponentInfo";
 import {Port}          from "../Port";
 
-import {BaseObjectImpl} from "./BaseObject";
-import {PortImpl}       from "./Port";
+import { BaseObjectImpl } from "./BaseObject";
+import { PortImpl } from "./Port";
```
LeonMontealegre, line 10: nit: remove spaces here

LeonMontealegre (changes requested):
Lots of nits and things that you can abuse the JS/TS syntax to make prettier

Also one big question/comment:

Can/should we move all this transaction logic within `CircuitInternal` into `HistoryManager`? Or some `TransactionManager`?

src/app/core/internal/impl/CircuitInternal.ts
```diff
@@ -1,72 +1,172 @@
...
+        let op: MultiOp | undefined = undefined;
+        if (this.transactionOps.length > 0) {
+            op = { kind: "MultiOp", ops: this.transactionOps };
+            this.history.push(op);
+            this.transactionOps = [];
+
+            // TODO: Emit events for changes...
+        }
+        this.transaction = false;
+        return op;
```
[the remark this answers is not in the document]
KevinMackenzie: `this.transaction = false` always needs to happen as the last step

src/app/core/internal/impl/CircuitInternal.ts
```diff
@@ -3,70 +3,177 @@ import {GUID} from "core/schema/GUID";
...
+    // Explicit variable only needed for asserts.  JS threading model means no mutex needed, but it
+    // is possible for an async call mid-transaction to yield control-flow to another task and break
+    // atomicity of transactions.  Any behavior that yields control-flow mid transaction is ILLEGAL,
+    // and thus guarded by asserts.
+    protected transaction: boolean;
+
+    // TODO: Can the emitted events be completely generated from this Ops list?
+    protected transactionOps: CircuitOp[];
+
+    public constructor(componentsTypes: ComponentInfoProvider) {
```
LeonMontealegre, line 37: nit: name `componentTypes`

src/app/core/internal/impl/CircuitInternal.ts
```diff
@@ -3,70 +3,177 @@ import {GUID} from "core/schema/GUID";
...
+            // TODO: Emit events for changes...
+        }
+        this.transaction = false;
+        return op;
+    }
+    public cancelTransaction(): void {
+        this.assertTransactionState(true);
+        if (this.transactionOps.length > 0) {
+            InvertMultiOp({ kind: "MultiOp", ops: this.transactionOps }).ops
+                .forEach((rollbackOp) => { this.applyOp(rollbackOp); });
```
LeonMontealegre, line 84: nit, just do
`.forEach((rollbackOp) => this.applyOp(rollbackOp))`

src/app/core/internal/impl/ComponentInfo.ts
```diff
@@ -0,0 +1,117 @@
...
+// TODO: seems kind of heavy
+export function IsValidPortList(info: ComponentInfo, ports: Schema.Port[]): boolean {
+    const countMap = new Map<string, number>();
+    ports.forEach(({ group }) =>
+        countMap.set(group, (countMap.get(group) ?? 0) + 1));
+
+    const counts: Record<string, number> = {};
+    info.portGroups.forEach((g) =>
+        { counts[g] = countMap.get(g) ?? -1});
+    return info.isValidPortConfig({ counts });
```
LeonMontealegre, line 38: Mmm I think a more straightforward way would be
```ts
// Get initial counts for each defined port group
const counts = info.portGroups.reduce((g, counts) => ({ ...counts, [g]: 0 }), {});
if (ports.some((group) => !(group in counts)))
    return false;
// Count the ports
ports.forEach(({ group }) => counts[group]++);
return info.isValidPortConfig({ counts });
```

src/app/core/internal/impl/CircuitInternal.ts
```diff
@@ -3,70 +3,177 @@ import {GUID} from "core/schema/GUID";
...
+        const oldVal = obj[key] as O["props"][K] | undefined;
+        this.applyOp({ kind: "SetPropertyOp", key, newVal: val, oldVal });
+    }
+
+    public setPortConfig(id: GUID, portConfig: PortConfig): void {
+        const obj = this.getCompByID(id);
+        if (!obj)
+            throw new Error("Failed to set port config: invalid GUID");
+        const componentInfo = this.componentTypes.get(obj.kind);
+        if (componentInfo === undefined)
```
LeonMontealegre, line 156: nit: can also `if (!componentInfo)` here

src/app/core/internal/impl/CircuitInternal.ts
```diff
@@ -3,70 +3,177 @@ import {GUID} from "core/schema/GUID";
...
+            throw new Error("Failed to set port config: Unknown component type (this is really bad)");
+        const ports = componentInfo.makePortsForConfig(id, portConfig);
+        if (ports === undefined)
```
LeonMontealegre, line 159: and here

LeonMontealegre (changes requested):
just a few small changes

src/app/core/internal/impl/ComponentInfo.ts
KevinMackenzie: i don't think the original or the new one does the right thing.  `counts` should contain all groups in `ports`, even if they don't exist in `info`.  `isValidPortConfig` already needs to accept arbitrary `PortConfig`s.  This does the thing I was intending:

```typescript
    // Count ports in each group
    const countMap = new Map<string, number>();
    ports.forEach(({ group }) =>
        countMap.set(group, (countMap.get(group) ?? 0) + 1));

    // Convert map to record
    const counts = {} as Record<string, number>;
    countMap.forEach((v, k) => counts[k] = v);

    return info.isValidPortConfig({ counts });
```

LeonMontealegre: Okay then I mean it's even easier:
```ts
const counts = {} as Record<string, number>;
ports.forEach(({ group }) =>
    (counts[group] = (counts[group ?? 0) + 1));
return info.isValidPortConfig({ counts });
```
The main point being that the map is unnecessary

LeonMontealegre (approved):

PR #1223, opened 2023-02-02
Began adding digital API overloads

LeonMontealegre:


src/app/core/public/api/impl/Circuit.ts
```diff
@@ -16,32 +17,29 @@ import {Obj}           from "../Obj";
...
-import {CircuitState}  from "./CircuitState";
-import {ComponentImpl} from "./Component";
-import {PortImpl}      from "./Port";
-import {WireImpl}      from "./Wire";
+import {CircuitState} from "./CircuitState";
 
 
-export abstract class CircuitImpl implements Circuit {
-    protected readonly state: CircuitState;
+export abstract class CircuitImpl implements Circuit, CircuitState {
```
KevinMackenzie, line 23: `implements Circuit` redundant here?

src/app/core/public/api/impl/Component.ts
```diff
@@ -1,45 +1,46 @@
...
+        const obj = this.internal.getCompByID(this.id);
         if (!obj)
             throw new Error(`API Component: Attempted to get component with ID ${this.id} could not find it!`);
         return obj;
     }
 
-    public get info(): ComponentInfo {
-        throw new Error("Unimplemented");
-    }
+    public abstract get info(): ReturnType<State["constructComponent"]>["info"];
```
KevinMackenzie, line 26: fancy

src/app/digital/public/api/DigitalPort.ts
```diff
@@ -0,0 +1,11 @@
+import {Port} from "core/public";
+
+import {DigitalComponent} from "./DigitalComponent";
+
+
+export interface DigitalPort extends Port {
+    readonly parent: DigitalComponent;
+
+    readonly isInputPort: boolean;
+    readonly isOutputPort: boolean;
```
KevinMackenzie, line 10: What's the thought behind `isXPort` vs something like `portDirection: "input" | "output"`?

src/app/core/public/api/impl/CircuitState.ts
```diff
@@ -2,13 +2,22 @@ import {CircuitInternal}   from "core/internal";
...
 
+import {Circuit}   from "../Circuit";
+import {Component} from "../Component";
+import {Port}      from "../Port";
+import {Wire}      from "../Wire";
+
 
 // Utility object to pass around the Circuit API state to the API Component/Port/etc
-export interface CircuitState {
+export interface CircuitState extends Circuit {
```
KevinMackenzie, line 12:
```typescript
export interface CircuitState<
  ComponentT extends Component = Component,
  WireT extends Wire = Wire,
  PortT extends Port = Port
> extends Circuit { 
  // ...
  constructComponent(id: string): ComponentT;
  // ...
}
```

Would something like this make
```typescript
interface DigitalCircuitState extends CircuitState<DigitalComponent, DigitalWire, DigitalPort>
```
possible and remove the dup'd fields/functions in that interface?  Would this make `CircuitImpl` a nightmare?

src/app/core/public/api/impl/Component.ts
```diff
@@ -48,14 +49,19 @@ export class ComponentImpl extends BaseObjectImpl implements Component {
...
-            .map((id) => new PortImpl(this.state, id));
+    public get ports(): Record<string, Port[]> {
+        return FromConcatenatedEntries(
+            [...this.internal.getPortsForComponent(this.id)]
+            .map((id) => this.circuit.constructPort(id))
+            .map((p) => [p.group, p])
+        );
     }
+
+    public abstract firstAvailable(group: string): Port | undefined;
```
KevinMackenzie, line 66: what's the reason for being abstract?  If ports have an `index` then this shouldn't depend on analog/digital, right?

KevinMackenzie (commented):
Seems reasonable.  There's a few things that could be consolidated but the high level makes sense.  `CircuitState` isn't an awesome name, but I don't have a better alternative.

src/app/core/public/api/impl/Circuit.ts
LeonMontealegre: It is, but idk I like the explicitness of making sure it's still a Circuit

src/app/core/public/api/impl/Component.ts
LeonMontealegre: I hate this tbh, it feels extremely hacky
The alternative I can think of is passing like a bunch of template arguments into Circuit/Component/Wire/etc
like:
`Component<DigitalCircuit, DigitalComponent, DigitalWire, DigitalPort>` to specify each sub-type of Circuit/Component/Wire/Port idk what do you think

src/app/digital/public/api/DigitalPort.ts
LeonMontealegre: Probably could have both

src/app/core/public/api/impl/CircuitState.ts
LeonMontealegre: I like this idea, let me give it a try

src/app/core/public/api/impl/Component.ts
LeonMontealegre: Because digital ports could be unavailable if they are input ports and already have a connection

KevinMackenzie: But that's still a problem with any kind of port.  Otherwise what's the meaning of `firstAvailable`?

src/app/core/public/api/impl/Component.ts
KevinMackenzie: Maybe not if there isn't another reason to?

src/app/core/public/api/impl/Component.ts
LeonMontealegre: `firstAvailable` would be the first "available to be connected to" port in the given group -- we don't currently have like a `isPortAvailable` function anywhere so I guess I thought I'd do it here, where should this behavior be described then?

KevinMackenzie: Maybe I'm forgetting a use-case, but should this be on `DigitalComponent` instead?  My understanding is that for analog it would always return the first port anyway.  Is this right?

LeonMontealegre: Mmm that's a good point actually yeah, it being unique to DigitalComponent makes sense

PR #1224, opened 2023-02-04
Error propagation proposal / integration.

KevinMackenzie:
- Adds rust-style `Option` and `Result` types for error propagation
    - Option is like `T | undefined` but with more FP features
    - Result is like Option, but can contain error info
- Uses JS `Error` throwing for irrecoverable errors.
- Converts CircuitInternal to new error conventions.
- Adjusts `ObjectInfo`/`ComponentInfo` interfaces to provide more useful `Result`s.

Changes outside of `impl` are mostly suggestions / to get the tests to run.

https://doc.rust-lang.org/std/result/
https://doc.rust-lang.org/std/option/


src/app/core/internal/impl/CircuitInternal.ts
```diff
@@ -370,38 +361,41 @@ export class CircuitInternal {
...
+            .andThen((newPorts) => {
+                // TODO: match old ports to new ports based on group/index so ID's / refs are reused.
+
+                // Collect old ports
+                const oldPorts = [] as Schema.Port[];
+                this.getPortsForComponent(id)
+                    .uponOk((ids) => ids.forEach((id) => {
+                        // Safe unwrap according to invariants
+                        oldPorts.push(this.getPortByID(id).unwrap());
+                    })).unwrap();
```
LeonMontealegre, line 387:
```ts
const oldPorts = [] as Schema.Port[];
this.getPortsForComponent(id)?.forEach((id) => {
  // Safe cast according to invariants
  oldPorts.push(this.getPortByID(id) as Schema.Port);
});
```

=>

```ts
const oldPorts = this.getPortsForComponent(id).map((id) => this.getPortByID(id));
```

src/app/core/internal/impl/CircuitInternal.ts
```diff
@@ -111,40 +123,30 @@ export class CircuitInternal {
...
-                checkWireConnectivity(port1, port2);
-                checkWireConnectivity(port2, port1);
-
-                (op.inverted ? deleteWire : addWire)(op.w);
-                break;
+                return this.getPortByID(op.w.p1)
+                    .andThen((port1) => this.getPortByID(op.w.p2)
+                        .andThen((port2) => checkWireConnectivity(port1, port2)
+                            .andThen((_) => checkWireConnectivity(port2, port1)
+                                .map((_) => { (op.inverted ? deleteWire : addWire)(op.w); }))));
```
LeonMontealegre, line 133: maybe consider a `.and` method

KevinMackenzie: Played around with a method like:
```ts
public andBind<U>(u: Result<U, E>): Result<[T, U], E>
```
but I don't think the current state of things warrants extending `Result`.  Punting this until we have cases that would benefit from this more (`N>=4`)

LeonMontealegre (approved):

PR #1226, opened 2023-02-07
Implemented selectionsMidpoint function and tests

rren8957:


src/app/core/public/api/impl/Circuit.ts
```diff
@@ -114,8 +115,18 @@ export abstract class CircuitImpl implements Circuit {
     public pickObjectRange(bounds: Rect): Obj[] {
         throw new Error("Unimplemented");
     }
+    
     public selectedObjs(): Obj[] {
-        throw new Error("Unimplemented");
+        const selectedIds = this.selections.get();
+        var selectedObjs = [];
```
TGCrystal, line 121: nit: Use `const`/`let` instead of `var`

src/app/core/public/api/impl/Circuit.ts
```diff
@@ -114,8 +115,18 @@ export abstract class CircuitImpl implements Circuit {
...
+
+        for (var id of selectedIds) {
+            var obj = this.getObj(id);
+            if (obj != undefined) {
```
TGCrystal, line 125: nit: use `!==` instead of `!=`

src/app/core/public/api/impl/Circuit.ts
```diff
@@ -155,7 +166,35 @@ export abstract class CircuitImpl implements Circuit {
         //  For now, ignore the `space`, and ignore any non-Component
         //   objects that are selected
         //  From these components, average their positions
-        throw new Error("Method not implemented.");
+
+        var allComponents = this.selections.get();
+
+        var xPosition = 0;
+        var yPosition = 0;
+        var comp = null;
```
TGCrystal, line 174: nit: Use `undefined` instead of `null`
In this case, you don't have to initialize `comp` as anything here. You can just create the variable when it is used below, like
```typescript
const comp = this.getComponent(allComponents[0]);
```
and
```typescript
const comp = this.getComponent(id);
```

src/app/core/public/api/impl/Circuit.ts
```diff
@@ -155,7 +166,35 @@ export abstract class CircuitImpl implements Circuit {
...
+        
+        // Case: only one component selected
+        if (allComponents.length == 1) {
```
TGCrystal, line 177: nit: Use `===` instead of `==`

TGCrystal (changes requested):

TGCrystal (approved):
Looks good to me! I added Leon as a reviewer, when he gets the chance he will look it over and approve or request changes

src/app/core/public/api/impl/Circuit.ts
```diff
@@ -155,7 +166,34 @@ export abstract class CircuitImpl implements Circuit {
...
+            const comp = this.getComponent(id);
+            if (comp !== undefined) {
+                xPosition += comp.pos.x;
+                yPosition += comp.pos.y;
+            }
+        }
+
+        // Calculate average position
+        xPosition = xPosition / allComponents.length;
+        yPosition = yPosition / allComponents.length;
```
LeonMontealegre, line 193: I'm going to have you attempt to condense this using `.map` and `.filter` as I showed above! \
(hint: take advantage of `comp.pos` being a `Vector` and don't use x/y separately)\
(hint 2: look into the `.reduce` function, which can take an array of things (like numbers/vector), and reduce them into a single number/vector)

src/app/core/public/api/impl/Circuit.ts
```diff
@@ -155,7 +166,34 @@ export abstract class CircuitImpl implements Circuit {
...
+
+        return V(xPosition, yPosition);
+
```
LeonMontealegre, line 196: nit: remove extra blank line

src/app/core/public/api/impl/Circuit.ts
```diff
@@ -155,7 +166,34 @@ export abstract class CircuitImpl implements Circuit {
         //  For now, ignore the `space`, and ignore any non-Component
         //   objects that are selected
         //  From these components, average their positions
-        throw new Error("Method not implemented.");
+
```
LeonMontealegre, line 169: nit: remove extra blank line

src/app/tests/SelectionsMidpoint.test.ts
```diff
@@ -0,0 +1,61 @@
+import {V} from "Vector";
+import {CreateCircuit} from "digital/public";
+
+import "./Extensions";
+
+
+describe("SelectionsMidpoint", () => {
+    test("Single Selection", () => {
+
```
LeonMontealegre, line 9: nit: remove extra blank lines above and below each code block

src/app/tests/SelectionsMidpoint.test.ts
```diff
@@ -0,0 +1,61 @@
...
+        expect(c1.isSelected).toEqual(true);
+        expect(l1.isSelected).toEqual(true);
+
+        // Calculate midpoint position using method
+        const methodMidpoint = circuit.selectionsMidpoint("screen");
+
+        // Calculate actual midpoint position
+        const xPosition = (s1.pos.x + s2.pos.x + c1.pos.x + l1.pos.x) / 4;
+        const yPosition = (s1.pos.y + s2.pos.y + c1.pos.y + l1.pos.y) / 4;
+        const actualMidpoint = V(xPosition, yPosition);
```
LeonMontealegre, line 56: Generally for unit tests, we want as little code as possible because we don't want bugs to exist in our tests (since we don't want to have tests for the tests)
So generally you don't want to redo complex operations in code, instead in your setup, use a setup that has a specific output that you already know.
I.e.
Set the positions of the components such that the midpoint is at an easy-to-know position, like with the LED at `V(6, 0)` which causes the midpoint to be at `V(-1 ,0)`, and then just set `actualMidpoint = V(-1, 0)`

src/app/tests/SelectionsMidpoint.test.ts
```diff
@@ -0,0 +1,61 @@
...
+        // Select created component
+        s1.isSelected = true;
+
+        // Check component has been selected
+        expect(s1.isSelected).toEqual(true);
+
+        // Calculate midpoint position using method
+        const sm1 = circuit.selectionsMidpoint("screen");
+        // Calculate actual midpoint position
+        const actualMidpoint = s1.pos;
```
LeonMontealegre, line 23: This is a great practice, but I think it's sufficient to just do
`expect(sm1).toEqual(s1.pos)`

Moreso, we should actually just do `expect(sm1).toEqual(V(0, 0))` since tests are really the place we *want* to be hardcoding numbers

LeonMontealegre (changes requested):
Looking great! Just a few formatting fixes and some "best practice" stuff for the OpenCircuits codebase!!

Another request, can you also please add a TSDoc-style comment onto `src/app/core/public/api/Circuit.ts` (not `impl`) above your `selectionsMidpoint` method?

The format is like:

```ts
/**
 * Description.
 * 
 * @param space Description of parameters.
 * @returns Description of return value.
 */
selectionsMidpoint(space: Vector.Spaces): Vector;
```

src/app/core/public/api/Circuit.ts
```diff
@@ -45,6 +45,15 @@ export interface Circuit {
     getObjs(): Obj[];
     getComponentInfo(kind: string): ComponentInfo | undefined;
 
+    /**
+     * Returns the average of the positions of the components selected
+     * as a Vector object.
+     * 
+     * @param {String} space defines the coordinate-space which can be
```
LeonMontealegre, line 52: You dont need the type for tsdoc comments since the type is already defined in the code

src/app/core/public/api/impl/BaseObject.ts
```diff
@@ -37,10 +37,14 @@ export abstract class BaseObjectImpl implements BaseObject {
     }
 
     public set isSelected(val: boolean) {
-        throw new Error("Unimplemented");
+        if (val === true) {
```
LeonMontealegre, line 40: nit: just do `if (val)` and also remove all the curly braces since these are 1-liner if/else statement blocks

src/app/core/public/api/impl/Circuit.ts
```diff
@@ -1,4 +1,5 @@
 import {Vector} from "Vector";
+import {V} from "Vector";
```
LeonMontealegre, line 2: nit: Do `import {Vector, V} from "Vector";`

src/app/core/public/api/impl/Circuit.ts
```diff
@@ -155,7 +159,23 @@ export abstract class CircuitImpl implements Circuit {
...
+
+        // Case: no components are selected
+        if (allComponents.length === 0) {
+            return V(0,0)
+        }
+
+        // Case: One or more components are selected
+        let comps =  allComponents.map((id) => this.getComponent(id))
+                    .filter((comp) => (comp !== undefined)) as Component[];
+        let sumPositions = comps.reduce((vec1, vec2) => V(vec1.x, vec1.y).add(vec2.pos.x, vec2?.pos.y), V(0,0));
```
LeonMontealegre, line 172: `Vector` handily already provides an easy way to just add two vectors, so you can just do
`(vec1, vec2) => vec1.add(vec2.pos)`

src/app/core/public/api/impl/Circuit.ts
```diff
@@ -155,7 +159,23 @@ export abstract class CircuitImpl implements Circuit {
         //  For now, ignore the `space`, and ignore any non-Component
         //   objects that are selected
         //  From these components, average their positions
-        throw new Error("Method not implemented.");
+        const allComponents = this.selections.get();
```
LeonMontealegre, line 162: I would get all the actual components up here so
```ts
const allComponents = this.selections.get()
    .map((id) => this.getComponent(id))
    .filter((comp) => (comp !== undefined)) as Component[];
```

src/app/core/public/api/impl/Circuit.ts
```diff
@@ -155,7 +159,23 @@ export abstract class CircuitImpl implements Circuit {
...
+
+        // Case: no components are selected
+        if (allComponents.length === 0) {
+            return V(0,0)
+        }
```
LeonMontealegre, line 167: nit: don't need the braces for single-line if-statements

src/app/core/public/api/impl/Circuit.ts
LeonMontealegre: I would also just 1-line this as:

```ts
const sumPosition = allComponents
    .map((c) => c.pos)
    .reduce((sum, v) => sum.add(v));
```

src/app/core/public/api/impl/Circuit.ts
```diff
@@ -155,7 +159,23 @@ export abstract class CircuitImpl implements Circuit {
...
+        // Case: One or more components are selected
+        let comps =  allComponents.map((id) => this.getComponent(id))
+                    .filter((comp) => (comp !== undefined)) as Component[];
+        let sumPositions = comps.reduce((vec1, vec2) => V(vec1.x, vec1.y).add(vec2.pos.x, vec2?.pos.y), V(0,0));
+
+        // Calculate average position
+        let xPosition = sumPositions.x / allComponents.length;
+        let yPosition = sumPositions.y / allComponents.length;
+
+        return V(xPosition, yPosition);
```
LeonMontealegre, line 178: And again `Vector` provides methods so you don't have to do things individually

simply `return sumPosition.scale(1 / allComponents.length);`

LeonMontealegre (changes requested):
Awesome job! Just a few more nits!

LeonMontealegre (approved):

PR #1232, opened 2023-03-03
Implemented firstAvailable() method and related test cases

rren8957:
Note: 2nd test case fails because connectWire() is not implemented

### DIFF
--- a/src/app/digital/public/api/DigitalComponent.ts
+++ b/src/app/digital/public/api/DigitalComponent.ts
@@ -10,8 +10,12 @@ export interface DigitalComponent extends Component {
      * Returns the first available port with the specified 
      * port group that a component contains.
      * 
+     * Note: output port => always available.
+     *       input port => must NOT be connected via a wire to another
+     *                     to be available.
+     * 
      * @param portGroup defines the desired port group.
      * @returns A Port if there is one available, undefined otherwise. 
      */
-    firstAvailable(portGroup: String): DigitalPort | undefined;
+    firstAvailable(portGroup: string): DigitalPort | undefined;
 }
--- a/src/app/digital/public/api/DigitalComponent.ts
+++ b/src/app/digital/public/api/DigitalComponent.ts
@@ -6,5 +6,12 @@ import {DigitalComponentInfo} from "./DigitalComponentInfo";
 export interface DigitalComponent extends Component {
     readonly info: DigitalComponentInfo;
 
-    firstAvailable(portGroup: string): DigitalPort | undefined;
+    /**
+     * Returns the first available port with the specified 
+     * port group that a component contains.
+     * 
+     * @param portGroup defines the desired port group.
+     * @returns A Port if there is one available, undefined otherwise. 
+     */
+    firstAvailable(portGroup: String): DigitalPort | undefined;
 }
--- a/src/app/digital/public/api/DigitalComponent.ts
+++ b/src/app/digital/public/api/DigitalComponent.ts
@@ -6,7 +6,7 @@ import {DigitalComponentInfo} from "./DigitalComponentInfo";
 export interface DigitalComponent extends Component {
     readonly info: DigitalComponentInfo;
 
-    firstAvailable(portGroup: string): DigitalPort;
+    firstAvailable(portGroup: string): DigitalPort | undefined;
 
     // public function firstAvailable(portGroup: string): DigitalPort {
     //     let inputPorts;
--- a/src/app/digital/public/api/DigitalComponent.ts
+++ b/src/app/digital/public/api/DigitalComponent.ts
@@ -7,16 +7,4 @@ export interface DigitalComponent extends Component {
     readonly info: DigitalComponentInfo;
 
     firstAvailable(portGroup: string): DigitalPort | undefined;
-
-    // public function firstAvailable(portGroup: string): DigitalPort {
-    //     let inputPorts;
-    //     if (portGroup === "input") 
-    //         inputPorts = DigitalComponentInfo.inputPortGroups;
-    //     else if (portGroup === "output") 
-    //         inputPorts = DigitalComponentInfo.inputPortGroups;
-        
-        
-    //         return DigitalPortImpl;
-    
-    // }
 }
--- a/src/app/digital/public/api/DigitalComponent.ts
+++ b/src/app/digital/public/api/DigitalComponent.ts
@@ -1,8 +1,22 @@
 import {Component} from "core/public";
 
+import {DigitalPort} from "./DigitalPort";
 import {DigitalComponentInfo} from "./DigitalComponentInfo";
-
 
 export interface DigitalComponent extends Component {
     readonly info: DigitalComponentInfo;
+
+    firstAvailable(portGroup: string): DigitalPort;
+
+    // public function firstAvailable(portGroup: string): DigitalPort {
+    //     let inputPorts;
+    //     if (portGroup === "input") 
+    //         inputPorts = DigitalComponentInfo.inputPortGroups;
+    //     else if (portGroup === "output") 
+    //         inputPorts = DigitalComponentInfo.inputPortGroups;
+        
+        
+    //         return DigitalPortImpl;
+    
+    // }
 }
--- a/src/app/digital/public/api/DigitalPort.ts
+++ b/src/app/digital/public/api/DigitalPort.ts
@@ -11,6 +11,10 @@ export interface DigitalPort extends Port {
 
     /**
      * Returns true if a port is available, false otherwise.
+     * Availabilty of ports:
+     *    - If it's an output port, it's always available.
+     *    - If it's an input port, to be considered available,
+     *      it must NOT be connected via a wire to another port.
      * 
      * @returns True or false.
      */
--- a/src/app/digital/public/api/DigitalPort.ts
+++ b/src/app/digital/public/api/DigitalPort.ts
@@ -8,4 +8,6 @@ export interface DigitalPort extends Port {
 
     readonly isInputPort: boolean;
     readonly isOutputPort: boolean;
+
+    isAvailable(): boolean;
 }
--- a/src/app/digital/public/api/DigitalPort.ts
+++ b/src/app/digital/public/api/DigitalPort.ts
@@ -9,5 +9,10 @@ export interface DigitalPort extends Port {
     readonly isInputPort: boolean;
     readonly isOutputPort: boolean;
 
+    /**
+     * Returns true if a port is available, false otherwise.
+     * 
+     * @returns True or false.
+     */
     isAvailable(): boolean;
 }
--- a/src/app/digital/public/api/impl/DigitalComponent.ts
+++ b/src/app/digital/public/api/impl/DigitalComponent.ts
@@ -7,6 +7,7 @@ import {DigitalWire}          from "../DigitalWire";
 
 import {DigitalCircuitState}      from "./DigitalCircuitState";
 import {DigitalComponentInfoImpl} from "./DigitalComponentInfo";
+import { DigitalPortImpl } from "./DigitalPort";
 
 
 export class DigitalComponentImpl extends ComponentImpl<
@@ -17,12 +18,27 @@ export class DigitalComponentImpl extends ComponentImpl<
         return new DigitalComponentInfoImpl(this.circuit, this.kind);
     }
 
-    public override firstAvailable(portGroup: string): DigitalPort {
-        let inputPorts;
-        if (portGroup === "input") 
-            inputPorts = this.info.inputPortGroups;
-        else if (portGroup === "output") 
-            inputPorts = this.info.inputPortGroups;
+    public override firstAvailable(portGroup: string): DigitalPort | undefined {
+        let Ports;
+
+        // Get all ports of the specified port group
+        if (portGroup === "input") {
+            Ports = this.info.inputPortGroups;
+        } else {
+            Ports = this.info.outputPortGroups;
+            // output ports are always available
+            return new DigitalPortImpl(this.circuit, Ports[0]);
+        }
+            
+        // Case: no "input" ports are available
+        if (Ports.length == 0) {
+            return undefined;
+        } 
+
+        for (var portNum of Ports) {
+            
+        }
+        return new DigitalPortImpl(this.circuit, Ports[0]);
     
     }
 }
--- a/src/app/digital/public/api/impl/DigitalComponent.ts
+++ b/src/app/digital/public/api/impl/DigitalComponent.ts
@@ -23,25 +23,25 @@ export class DigitalComponentImpl extends ComponentImpl<
         if (!this.info.portGroups.includes(portGroup))
             return undefined; // Invalid port group for the component
         
-        const ports = this.internal.getPortsForComponent(this.id);
+        const ports = this.internal.getPortsForComponent(this.id).unwrap();
 
         // Find out if the portGroup is of type input or output
         const isInputType = this.info.inputPortGroups.includes(portGroup);
         const isOutputType = this.info.outputPortGroups.includes(portGroup);
 
         const firstAvailableHelper = (id: string) => {
-            const portObject = this.internal.getPortByID(id)!;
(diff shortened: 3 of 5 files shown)


src/app/digital/public/api/impl/DigitalComponent.ts
```diff
@@ -1,4 +1,6 @@
+import { CircuitInternal } from "core/internal";
 import {ComponentImpl} from "core/public/api/impl/Component";
+import { exit } from "yargs";
```
LeonMontealegre, line 3: ?

src/app/digital/public/api/impl/DigitalComponent.ts
```diff
@@ -1,4 +1,6 @@
+import { CircuitInternal } from "core/internal";
```
LeonMontealegre, line 1: nit: remove spaces in import, this should be handled by ESLint, definitely install the VSCode Eslint plugin if you haven't already

src/app/digital/public/api/impl/DigitalComponent.ts
```diff
@@ -7,6 +9,7 @@ import {DigitalWire}          from "../DigitalWire";
 
 import {DigitalCircuitState}      from "./DigitalCircuitState";
 import {DigitalComponentInfoImpl} from "./DigitalComponentInfo";
+import { DigitalPortImpl } from "./DigitalPort";
```
LeonMontealegre, line 12: nit: spacing, use `yarn lint` or the ESLint plugin

src/app/digital/public/api/impl/DigitalComponent.ts
```diff
@@ -16,4 +19,38 @@ export class DigitalComponentImpl extends ComponentImpl<
     public override get info(): DigitalComponentInfo {
         return new DigitalComponentInfoImpl(this.circuit, this.kind);
     }
+
+    public override firstAvailable(portGroup: string): DigitalPort | undefined {
+        const Ports = this.internal.getPortsForComponent(this.id);
```
LeonMontealegre, line 24: nit: call it `ports`

src/app/digital/public/api/impl/DigitalComponent.ts
```diff
@@ -16,4 +19,38 @@ export class DigitalComponentImpl extends ComponentImpl<
...
+
+        if (portGroup === "input") {
```
LeonMontealegre, line 26: So `portGroup` is going to be the actual `port.group`, not "input" or "output",

i.e. a Multiplexer will have the groups: "inputs", "outputs", "selects"

"inputs" and "selects" are both "Input Ports" and will both be in `inputPortGroups`
"outputs" are "Output Ports" and will be in `outputPortGroups`

the user should be able to do `myMultiplexer.firstAvailable("selects")`
which will return a different port then `myMultiplexer.firstAvailable("inputs")`

src/app/digital/public/api/impl/DigitalComponent.ts
```diff
@@ -16,4 +19,38 @@ export class DigitalComponentImpl extends ComponentImpl<
...
+                return undefined;
+            } else {
+                for(var p of Ports.values()) {
+                    let portObject = this.internal.getPortByID(p);
+                    let portWire = this.internal.getWiresForPort(p); // no wires = available
+                    if (portObject !== undefined && portObject.group === 'outputs') {
+                        return new DigitalPortImpl(this.circuit, p);  
+                    }
+                }
+            }
```
LeonMontealegre, line 53: These code blocks are very similar, try and see if you can condense them

Also note that you can make use of `.find` to avoid writing a for loop explicitly

src/app/digital/public/api/impl/DigitalPort.ts
```diff
@@ -20,4 +20,17 @@ export class DigitalPortImpl extends PortImpl<
...
+    public isAvailable(): boolean {
+        if (this.isOutputPort) { // output port case
+            return true;
+        } else { // input port case 
+            const portWire = this.internal.getWiresForPort(this.id);
+            if (portWire.size === 0)
+                return true;
+        }
+        
+        return false
```
LeonMontealegre, line 34: Nice! and we can write this more concisely:

```ts
// Output ports are always available for more connections
if (this.isOutputPort)
    return true;
// Input ports are only available if there isn't a connection already
const wires = this.internal.getWiresForPort(this.id);
return (wires.size === 0);
```

LeonMontealegre (changes requested):
Make sure to install an ESLint plugin so you can see linting errors, alternatively you can run `yarn lint` but it's going to complain about a bunch of other files too, so you'll have to look for the output only for your files (but prefer install an ESLint extension)

src/app/digital/public/api/DigitalComponent.ts
```diff
@@ -1,8 +1,10 @@
 import {Component} from "core/public";
 
+import {DigitalPort} from "./DigitalPort";
 import {DigitalComponentInfo} from "./DigitalComponentInfo";
 
-
 export interface DigitalComponent extends Component {
     readonly info: DigitalComponentInfo;
+
+    firstAvailable(portGroup: string): DigitalPort | undefined;
```
LeonMontealegre, line 9: Can you add a TSDoc-style comment above this explaining what it does?

src/app/digital/public/api/DigitalPort.ts
```diff
@@ -8,4 +8,6 @@ export interface DigitalPort extends Port {
 
     readonly isInputPort: boolean;
     readonly isOutputPort: boolean;
+
+    isAvailable(): boolean;
```
LeonMontealegre, line 12: Can you add a TSDoc-style comment above this explaining what it does?

src/app/digital/public/api/impl/DigitalPort.ts
```diff
@@ -20,4 +20,15 @@ export class DigitalPortImpl extends PortImpl<
         this.isInputPort  =  (this.parent.info.inputPortGroups.includes(this.group));
         this.isOutputPort = (this.parent.info.outputPortGroups.includes(this.group));
     }
+
+    // returns true if a port is available, false otherwise
+    public isAvailable(): boolean {
+        // Output ports are always available for more connections
+        if (this.isOutputPort)
+        return true;
```
LeonMontealegre, line 28: nit: indent

src/app/tests/FirstAvailable.test.ts
```diff
@@ -0,0 +1,33 @@
...
+        expect(c.firstAvailable("inputs")?.id).toEqual(c.ports["inputs"][0].id);
+        expect(c.firstAvailable("selects")?.id).toEqual(c.ports["selects"][0].id);
+    });
+    test("Only some/no ports are available", () => {
+        const circuit = CreateCircuit();
+
+        const s1 = circuit.placeComponentAt(V(-5,  5), "Switch");
+        const c1 = circuit.placeComponentAt(V(0,   0), "ANDGate");
+
+        circuit.connectWire(s1.ports["outputs"][0], c1.ports["inputs"][0])
```
LeonMontealegre, line 25: nit: semi-colon at end

src/app/digital/public/api/impl/DigitalComponent.ts
```diff
@@ -16,4 +18,34 @@ export class DigitalComponentImpl extends ComponentImpl<
...
+        const inputTypes = this.info.inputPortGroups;
+        const outputTypes = this.info.outputPortGroups;
+
+        const isInputType = inputTypes.includes(portGroup);
+        const isOutputType = outputTypes.includes(portGroup)
+
+        if (!isInputType && !isOutputType) 
+            return undefined;
+
+        function firstAvailableHelper(comp: DigitalComponentImpl, id: string): Boolean {
```
LeonMontealegre, line 34: declare this as
```ts
const firstAvailableHelper = (id: string) => {
    ...
}
```
and then you can just use `this` and not pass in `comp` explicitly

src/app/digital/public/api/impl/DigitalComponent.ts
```diff
@@ -16,4 +18,34 @@ export class DigitalComponentImpl extends ComponentImpl<
...
+    
+            if (comp.info.inputPortGroups && portObject !== undefined && portObject.group === portGroup)
+                return true;  
+            if (comp.info.outputPortGroups && portObject !== undefined && portObject.group === portGroup && portWire.size === 0) {
+                return true;  
+            }
+            return false;
+        }
+
+        const match = Array.from(ports.values()).find(p => firstAvailableHelper(this, p));
```
LeonMontealegre, line 46: make changes to `firstAvailableHelper` then you can just do
`[...ports].find(firstAvailableHelper);`

src/app/digital/public/api/impl/DigitalComponent.ts
```diff
@@ -16,4 +18,34 @@ export class DigitalComponentImpl extends ComponentImpl<
...
+        const isOutputType = outputTypes.includes(portGroup)
+
+        if (!isInputType && !isOutputType) 
+            return undefined;
+
+        function firstAvailableHelper(comp: DigitalComponentImpl, id: string): Boolean {
+            let portObject = comp.internal.getPortByID(id);
+            let portWire = comp.internal.getWiresForPort(id); // no wires = available
+    
+            if (comp.info.inputPortGroups && portObject !== undefined && portObject.group === portGroup)
```
LeonMontealegre, line 38: Checking `if (comp.info.inputPortGroups)` would just check if the array is defined or not but it should always be defined so I think it's a no-op currently

if you're trying to see if it's *in` the inputPortGroups, you gotta do `.includes(portObject.group)`

LeonMontealegre, line 38: instead of checking `if (portObject !== undefined)` both times, above this do
```ts
if (!portObject)
    return false;
```
But I think you can safely assume it will always be defined, so up at `portObject` so do
`const portObject = comp.internal.getPortByID(id)!;` (note the `!` which asserts that it's defined)

src/app/digital/public/api/impl/DigitalComponent.ts
```diff
@@ -16,4 +18,34 @@ export class DigitalComponentImpl extends ComponentImpl<
...
+
+        const isInputType = inputTypes.includes(portGroup);
+        const isOutputType = outputTypes.includes(portGroup)
+
+        if (!isInputType && !isOutputType) 
+            return undefined;
+
+        function firstAvailableHelper(comp: DigitalComponentImpl, id: string): Boolean {
+            let portObject = comp.internal.getPortByID(id);
+            let portWire = comp.internal.getWiresForPort(id); // no wires = available
```
LeonMontealegre, line 36: nit: these can both be `const` as they aren't changing

src/app/digital/public/api/impl/DigitalComponent.ts
```diff
@@ -16,4 +18,34 @@ export class DigitalComponentImpl extends ComponentImpl<
...
+        const ports = this.internal.getPortsForComponent(this.id);
+
+        const inputTypes = this.info.inputPortGroups;
+        const outputTypes = this.info.outputPortGroups;
+
+        const isInputType = inputTypes.includes(portGroup);
+        const isOutputType = outputTypes.includes(portGroup)
+
+        if (!isInputType && !isOutputType) 
+            return undefined;
```
LeonMontealegre, line 32: A valid portGroup will always be an input or output group (or we have a bigger problem)
So you can simplify this by just doing
```ts
if (!this.info.portGroups.includes(portGroup))
    return undefined; // Invalid port group for the component
```

LeonMontealegre (changes requested):
Looking good! Just a few formatting nits and some other stuff

src/app/digital/public/api/DigitalComponent.ts
```diff
@@ -1,8 +1,17 @@
...
     readonly info: DigitalComponentInfo;
+
+    /**
+     * Returns the first available port with the specified 
+     * port group that a component contains.
+     * 
+     * @param portGroup defines the desired port group.
+     * @returns A Port if there is one available, undefined otherwise. 
+     */
+    firstAvailable(portGroup: String): DigitalPort | undefined;
```
LeonMontealegre, line 16: nit: use `string` not `String`

src/app/digital/public/api/DigitalPort.ts
```diff
@@ -8,4 +8,11 @@ export interface DigitalPort extends Port {
 
     readonly isInputPort: boolean;
     readonly isOutputPort: boolean;
+
+    /**
+     * Returns true if a port is available, false otherwise.
```
LeonMontealegre, line 13: Let's be more specific about what "available" means, i.e. explain the difference b/t input and output ports and what makes them "available"

src/app/digital/public/api/impl/DigitalComponent.ts
```diff
@@ -16,4 +18,32 @@ export class DigitalComponentImpl extends ComponentImpl<
...
+                return true;  
+        
+            return false;
+        }
+
+        const match = [...ports].find(firstAvailableHelper);
+        if (match !== undefined)
+            return new DigitalPortImpl(this.circuit, match);  
+        return undefined;
+    }
```
LeonMontealegre, line 48: nit: let's invert the conditions here, i.e.
```ts
if (!match)
    return undefined;
return new DigitalPortImpl(this.circuit, match);
```

LeonMontealegre (changes requested):
Almost there! Just a few more nits

Also are you up-to-date with `model_refactor_api`? We recently merged an important change to `CircuitInternal` that might cause some changes in your code to take place too

src/app/digital/public/api/impl/DigitalComponent.ts
```diff
@@ -16,4 +18,32 @@ export class DigitalComponentImpl extends ComponentImpl<
...
+            return undefined; // Invalid port group for the component
+        
+        const ports = this.internal.getPortsForComponent(this.id);
+
+        // Find out if the portGroup is of type input or output
+        const isInputType = this.info.inputPortGroups.includes(portGroup);
+        const isOutputType = this.info.outputPortGroups.includes(portGroup);
+
+        const firstAvailableHelper = (id: string) => {
+            const portObject = this.internal.getPortByID(id)!;
```
LeonMontealegre, line 33: nit: do the `unwrap` up here, just `const portObject = this.internal.getPortByID(id).unwrap();` (you can remove the exclamation point now)

src/app/digital/public/api/impl/DigitalComponent.ts
```diff
@@ -16,4 +18,32 @@ export class DigitalComponentImpl extends ComponentImpl<
...
+            const portWire = this.internal.getWiresForPort(id); // no wires = available
```
LeonMontealegre, line 34: nit: unwrap here too

src/app/digital/public/api/impl/DigitalComponent.ts
```diff
@@ -16,4 +18,32 @@ export class DigitalComponentImpl extends ComponentImpl<
...
+
+            if (isInputType && portObject.unwrap().group === portGroup)
+                return true;  
+            if (isOutputType && portObject.unwrap().group === portGroup && portWire.unwrap().size === 0)
+                return true;  
+        
+            return false;
+        }
+
+        const match = [...ports.unwrap()].find(firstAvailableHelper)!;
```
LeonMontealegre, line 44: nit: unwrap up where you get `ports`

src/app/digital/public/api/impl/DigitalPort.ts
```diff
@@ -20,4 +20,15 @@ export class DigitalPortImpl extends PortImpl<
...
+
+    // returns true if a port is available, false otherwise
+    public isAvailable(): boolean {
+        // Output ports are always available for more connections
+        if (this.isOutputPort)
+            return true;
+
+        // Input ports are only available if there isn't a connection already
+        const wires = this.internal.getWiresForPort(this.id);
+        return (wires.unwrap().size === 0);
```
LeonMontealegre, line 32: nit: `unwrap` above

LeonMontealegre (changes requested):
lookin great! just a few last nits w/ `unwrap`

LeonMontealegre (approved):

PR #1236, opened 2023-03-21
Added Shortcuts to Analog

julianchen215:
We have succesfully added keyboard shortcuts to the Analog version of OpenCircuits:
“V” for voltage source
“G” for ground
“R” for Resistor
“C” for capacitor
“L” for inductor
“I” for current source


src/site/shared/containers/ItemNav/index.tsx
```diff
@@ -94,6 +99,24 @@ export const ItemNav = <D,>({ info, config, additionalData, getImgSrc, onDelete,
         if (info.currentlyPressedObject)
             setCurPressedObj(info.currentlyPressedObject);
     });
+    // Loops through the shorcuts 2d array in AnalogItemNav to see if any shortcuts
+    useDocEvent("keydown", (ev) => {
+        // Loop through each of the input shortcuts
+        for (var short of shortcuts){
```
LeonMontealegre, line 105: Never use var, use `let` or `const`, (`const` here)

LeonMontealegre (changes requested):
Just a quick change for now, will make a more thorough review later once you've fixed all of the linting errors (which you should please do)

src/site/shared/containers/ItemNav/index.tsx
```diff
@@ -94,6 +99,24 @@ export const ItemNav = <D,>({ info, config, additionalData, getImgSrc, onDelete,
...
+            if (ev.key === short[0]){
```
TGCrystal, line 106: Add space between `)` and `{`

src/site/shared/containers/ItemNav/index.tsx
```diff
@@ -118,6 +141,17 @@ export const ItemNav = <D,>({ info, config, additionalData, getImgSrc, onDelete,
 
     // Drop the current item on click (or on touch end)
     useDocEvent("click", (ev) => {
+        // If keyboard shorcut used to bring up component then allow to drop until "Esc" pressed
+        if(shortcut_flag){
```
TGCrystal, line 145:
```suggestion
        if (shortcut_flag) {
```

TGCrystal (changes requested):
Functionality wise, I think pressing the same shortcut for whatever item is currently selected (like if you press `s` after you already did once to select Switch) then it should unselect it just like right clicking does.

Bugs wise, the only issues I found relate to dragging an item from the item nav while a shortcut is being used
1. Open itemnav
2. Press s
3. right click
4. Click Segment Display
5. Click canvas to place Segment Display

It still tries to display a preview to place (it shouldn't), and it isn't finding an image to preview so it displays the alt placeholder text. Strangely enough, it does find a preview if you use Clock instead of Segment Display, although it still shouldn't have a preview.

A similar issue happens when:
1. Open itemnav
2. Press s
3. Drag an Segment Display onto the itemnav

The preview is still there (and still fails to display).

Aside from that and the fixes proposed by the linter and Leon, this looks good! I do like this new functionality quite a bit, it feels much more ergonomic and useful than I expected.

PR #1238, opened 2023-03-30
Model refactor api expr to circ

### LINKED ISSUE #1264: Reimplement Boolean Expression To Circuit Generator
# Reimplement Boolean Expression To Circuit Generator

Needs to be update to use the new API.

- [x] Use new API for creating/connecting within tree to circuit conversion
- [x] Use Result for error handling in core tree to circuit algorithm
- [x] Incorporate input validation to Result system
- [x] Make sure tests compile

This issue does not include wiring up the frontend.

### LINKED ISSUE #1273: Add `.toBeOk` jest matcher
For ease of testing `Result`s, we should add a `.toBeOk()` matcher. It can work like:
```typescript
expect(result).toBeOk()
```
and in the case of a failure, print/return/display the error(s).

This can be done now with
```typescript
expect(result.ok).toBeTruthy()
```
but that gives an almost entirely useless error if is not ok.

### LINKED ISSUE #1277: Add `.toIncludeError` jest matcher
To improve testing capabilities with `Result`, there should be a matcher to check if it has an error that contains the given error message.

This would be implemented by looping over the error messages and using [includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) on their error message to check for a match.

I initially planned on calling this `.toHaveError`, but I think `.toIncludeError` makes more sense to reflect the underlying usage of `.include()`

TGCrystal:
closes #1264 
closes #1273 
closes #1277


src/app/tests/Extensions.ts
```diff
@@ -81,6 +82,60 @@ expect.extend({
         };
     },
 
+    toBeOk(received: unknown) {
+        if (typeof received !== "object" || !received) {
```
LeonMontealegre, line 86: This is just not worth the effort lol, it's not like we're publishing this library, just do
```ts
if (!received || typeof received !== "object" || !("ok" in received)) {
  return {
    message: () => "supplied value is not a Result",
    pass: false,
  };
}
const result = received as Result;
...
```

src/site/pages/digital/src/utils/ComponentOrganizers.ts
LeonMontealegre, line 1: Why add this right now? `IOObjectSet` isn't a thing anymore anyways, so this doesn't even compile

LeonMontealegre (changes requested):
Only like a half-review for now:
can you please tell me what files I don't really have to pay attention to? i.e. which files are exact copy-and-pastes? Specifically the huge files like `GenerateInputTree`

TGCrystal:
> Only like a half-review for now: can you please tell me what files I don't really have to pay attention to? i.e. which files are exact copy-and-pastes? Specifically the huge files like `GenerateInputTree`

StringUtils, Formats, and DataStructures are the only things unchanged. Core algorithm files were changed to accommodate Result.

src/site/pages/digital/src/utils/ComponentOrganizers.ts
TGCrystal: Just so I don't have to hunt it down later. I'll comment it out for now

src/app/tests/Extensions.ts
TGCrystal: Well I did already put in the effort though but I can simplify it

src/app/core/utils/math/Graph.ts
```diff
@@ -179,3 +183,18 @@ export class Graph<V, E> {
     }
 
 }
+
+export function CreateGraph(circuit: Circuit): Graph<Component, Wire> {
```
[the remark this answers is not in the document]
TGCrystal: To what file?

src/site/pages/digital/src/utils/ExpressionParser/GenerateInputTree.ts
```diff
@@ -0,0 +1,272 @@
...
+        index = leftRet.index;
+        // If this isn't the right operation to apply, return
+        if (index >= tokens.length || tokens[index].type !== currentOp)
+            return Ok(leftRet);
+
+        index += 1;
+        if (index >= tokens.length)
+            return ErrE(`Missing Right Operand: "${ops[currentOp]}"`);
+        return generateInputTreeCore(tokens, ops, currentOpNum, index)
+                .andThen((rightRet): Result<NewTreeRetValue> => {
```
KevinMackenzie, line 140: You can use `map` here since this always returns `Ok(...)`

src/site/pages/digital/src/utils/ExpressionParser/GenerateInputTree.ts
```diff
@@ -0,0 +1,272 @@
...
+    if (rightToken.type === "!" || rightToken.type === "(") { // This case applies when the next token is "!" or "("
+        rightRet = generateInputTreeCore(tokens, ops, currentOpNum, index);
+    }
+    else if (rightToken.type === "input") { // This case would apply when an input follows a "!"
+        rightRet = Ok({ index: index+1, tree: { kind: "leaf", ident: rightToken.name } });
+    }
+    else {
+        return ErrE(`Invalid token "${ops[rightToken.type]}" following "${ops["!"]}"`);
+    }
+    return rightRet.andThen((rightRetVal) => {
```
KevinMackenzie, line 99: This can be `map`

src/site/pages/digital/src/utils/ExpressionParser/TreeToCircuit.ts
```diff
@@ -0,0 +1,130 @@
...
+
+/**
+ * Connects the first available output port on `prevComp` to the `newNode` on `newComp`.
+ *
+ * @param prevComp The component to connect the output of.
+ * @param newNode  First available `inputs` port on `newComp`.
+ * @param newComp  The component to connect the input of, only used for error message and return value.
+ * @returns        In the case of success, a `Result` wrapping `newComp`. An error in the following cases:
+ *                 - No output port found on `prevComp`.
+ *                 - Wire failed to create connecting `prevComp` to `newNode`.
```
KevinMackenzie, line 36: Can these failure cases ever occur?  It's preferable to throw exceptions and use `.unwrap()` in these cases if the `Err` path can only be triggered by internal bugs (this comment applies generally to this file).

TGCrystal: Yeah I got a bit overzealous with Result, I don't think it actually applies to this step. None of the error paths here should be triggered by user input, only by something going wrong internally.

src/app/core/public/api/CircuitUtils.ts
```diff
@@ -0,0 +1,24 @@
+import {Graph}   from "math/Graph";
```
LeonMontealegre, line 1: let's move this file into `src/shared/utils` I think

src/app/tests/Extensions.ts
```diff
@@ -81,40 +84,52 @@ expect.extend({
...
-    //             if (connections.includes(target))
-    //                 return true;
-
-    //             queue.push(...connections);
-    //         }
-    //         return bfs(queue, depth-1);
-    //     }
-
-    //     const pass = bfs([source], depth);
+    toBeOk(received: unknown) {
```
LeonMontealegre, line 87: You can just say `received: Result` and then you don't have to type-check it, I should probably do that for the others but I was probably referencing some js

src/app/tests/Extensions.ts
```diff
@@ -81,40 +84,52 @@ expect.extend({
...
+        }
+    },
 
-    //     return {
-    //         message: () => `expected ${source.getName()} to ${pass ? "" : "not "}be connected to ${target.getName()}` +
-    //                        ` within ${options.depth} connections`,
-    //         pass,
-    //     };
-    // },
+    toIncludeError(received: unknown, message: string) {
```
LeonMontealegre, line 111: same here

src/site/pages/digital/src/utils/ExpressionParser/TreeToCircuit.ts
```diff
@@ -0,0 +1,126 @@
...
+ * @throws When one of the leaf nodes of the InputTree references an input that is not inputs.
+ * @throws If any connections fail.
+ * @see TreeToCircuit
+ */
+function treeToCircuitCore(node: InputTree, inputs: Map<string, DigitalComponent>, circuit: DigitalCircuit):
+    DigitalComponent {
+    if (node.kind === "leaf") {
+        const input = inputs.get(node.ident);
+        if (!input)
+            throw new Error("Input Not Found: \"" + node.ident + "\"");
```
LeonMontealegre, line 66: Result/ErrE for this?

src/site/pages/digital/src/utils/ExpressionParser/TreeToCircuit.ts
```diff
@@ -0,0 +1,126 @@
...
+ * @param tree   The root node of the InputTree to convert or an empty array if tree is null.
+ * @param inputs The input components used by this expression.
+ * @param output The component that the circuit outputs to.
+ * @returns      The components and wires converted from the tree.
+ * @throws If `output` indicates a component without an input port.
+ * @throws When one of the leaf nodes of the InputTree references an input that is not inputs.
+ * @throws If any connections fail.
+ */
+export function TreeToCircuit(tree: InputTree, inputs: ReadonlyMap<string, string>,
+                              output: string): DigitalCircuit {
```
LeonMontealegre, line 106: Result this whole thing and all sub-functions?

src/site/pages/digital/src/utils/ComponentOrganizers.ts
```diff
@@ -0,0 +1,43 @@
+// TODO[model_refactor_api](trevor): Get this working for expression to circuit frontend
```
LeonMontealegre, line 1: Why not get this working now ?

src/site/pages/digital/tsconfig.json
```diff
@@ -12,6 +12,7 @@
             "shared/*":       ["../../shared/*"],
         }
     },
-    "include": ["../../shared/utils/types"],
+    // TODO[model_refactor_api](Leon) - Add back "src" to include
+    "include": ["src/utils/ExpressionParser", "../../shared/utils/types", "tests/utils/ExpressionParser.test.ts"],
```
LeonMontealegre, line 16: Why do you need to add these files?

LeonMontealegre (changes requested):

src/site/pages/digital/tsconfig.json
TGCrystal: So test can be run. It will have to be changed anyways before the main model refactor pr is merged so I just left it as is

LeonMontealegre (approved):

PR #1247, opened 2023-04-01
[1] Package upgrades and initial input changes + tools

LeonMontealegre:
Moved to @reduxjs/toolkit and use React 18 `createRoot` finally.

Also refactored the `InputManager` a bit to hold an `InputManagerState` that is passed downstream through the events so that subscribers can access the input state without needing a reference to `InputManager`.

Also added initial definitions for Tools and Handlers (please lmk what you think)


KevinMackenzie:
I'm not sure I understand why there needs to be `Tool`s and `ToolHandler`s.  Why the split?    Its not obvious to me when a behavior would be implemented as a `Tool` vs as a `ToolHandler` on the `DefaultTool`.  Would `ToolHandler`s ever be used outside of the `DefaultTool`?

LeonMontealegre:
Tools takeover control of the flow when activated and therefore generally last multiple frames/events and perform a single transaction over this time. 

ToolHandlers are basically just EventHandlers and only execute as a response to an event without ever taking control and are for instantaneous committed actions like CTRL+S to save. ZoomHandler is a particularly unique one since it does kinda do multiple things which it can do since none of the actions are committed since it's just a camera edit so maybe shouldn't be one and instead directly be in the DefaultTool but idk.

And yeah currently ToolHandlers only work on the DefaultTool. 

I'm open to a better/more descriptive name if you have one

src/app/core/public/api/impl/Circuit.ts
```diff
@@ -273,19 +276,22 @@ export abstract class CircuitImpl<
     public resize(w: number, h: number): void {
         this.view.resize(w, h);
     }
+    public get canvas() {
+        return this.view.getCanvas();
+    }
     public attachCanvas(canvas: HTMLCanvasElement): () => void {
         this.view.setCanvas(canvas);
+        // TODO: Figure out this event type more concretely
```
TGCrystal, line 284: TODO branch/name formatting

src/site/shared/tools/handlers/ToolHandler.ts
```diff
@@ -0,0 +1,24 @@
+import {Circuit}           from "core/public";
+import {InputManagerEvent} from "shared/utils/input/InputManagerEvent";
+
+
+export enum ToolHandlerResponse {
+    PASSTHROUGH = 0,
+    STOP = 1,
+}
+
+export interface ToolHandler {
```
KevinMackenzie, line 10: Maybe move to `DefaultTool.ts` so its more clear that's the exclusive user of this type, and to namespace it under `import {ToolHandler} from Default.ts`

src/site/shared/tools/handlers/ToolHandler.ts
```diff
@@ -0,0 +1,24 @@
+import {Circuit}           from "core/public";
+import {InputManagerEvent} from "shared/utils/input/InputManagerEvent";
+
+
+export enum ToolHandlerResponse {
+    PASSTHROUGH = 0,
```
KevinMackenzie, line 6: denote what these enums mean in the rest of the system here.

src/site/shared/utils/input/InputManager.ts
```diff
@@ -10,34 +10,59 @@ import {CalculateMidpoint} from "math/MathUtils";
...
-    pos: Vector;
-}
-export type CopyPasteInputEvent = {
-    type: "paste" | "copy" | "cut";
-    ev: ClipboardEvent;
-}
-export type OtherInputEvent = {
-    type: "mouseenter" | "mousemove" | "mouseleave" | "contextmenu" | "unknown";
-}
+export class InputManagerStateImpl implements InputManagerState {
```
KevinMackenzie, line 18: `UserInteractionState`?

src/site/shared/utils/input/InputManager.ts
```diff
@@ -49,28 +74,7 @@ export class InputManager extends Observable<InputManagerEvent> {
     /** The canvas the user is performing inputs on. */
```
KevinMackenzie, line 74: Maybe rename `InputManager` to `CanvasEventAdapter`/`InputEventAdapter`, or something.

src/site/shared/tools/handlers/ToolHandler.ts
LeonMontealegre: Will take care of this in https://github.com/OpenCircuits/OpenCircuits/pull/1248

TGCrystal (approved):

src/site/shared/tools/handlers/ToolHandler.ts
```diff
@@ -1,9 +1,11 @@
 import {Circuit}           from "core/public";
-import {InputManagerEvent} from "shared/utils/input/InputManagerEvent";
+import {InputAdapterEvent} from "shared/utils/input/InputAdapterEvent";
 
 
 export enum ToolHandlerResponse {
+    /** This response represents that the handler is okay with following handlers to execute in this event cycle. */
     PASSTHROUGH = 0,
+    /** This response represents that the handler does not want any other handlers to execute in this event cycle. */
```
KevinMackenzie, line 8: `other`->`subsequent`

KevinMackenzie (approved):
LGTM, only reviewed new commit since discussion wednesday

PR #1248, opened 2023-04-06
[2] Added CircuitDesigner initial API and misc tool/handler changes

LeonMontealegre:
Initial implementation of `CircuitDesigner` which will be an API in addition to `Circuit` that also manages the tools and input adapter.

Adds TranslateTool and SelectionHandler.

Also removes `space` from `Vector`.


src/site/shared/tools/handlers/ToolHandler.ts
```diff
@@ -1,26 +1,26 @@
...
 import {InputAdapterEvent} from "shared/utils/input/InputAdapterEvent";
 
 
 export enum ToolHandlerResponse {
     /** This response represents that the handler is okay with following handlers to execute in this event cycle. */
-    PASSTHROUGH = 0,
+    PASS = 0,
+    HALT = 1,
     /** This response represents that the handler does not want any other handlers to execute in this event cycle. */
-    STOP = 1,
```
KevinMackenzie, line 9: I think the ordering of comment/identifier got messed up in a merge/rebase

src/site/shared/utils/SnapUtils.ts
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Utility used for calculating positions of ports and wires when a component
+ * gets moved.
+ *
+ * @param oPos The origin position to snap.
+ * @param comp The component to consider snapping the connections for.
+ * @returns    The snapped position for `obj`.
+ */
+export function SnapToConnections(oPos: Vector, comp: Component): Vector {
```
KevinMackenzie, line 15: Consider a more descriptive name than `oPos`.  The doc comment also isn't very helpful in understanding what it means.

src/site/shared/utils/hooks/useDesigner.ts
```diff
@@ -0,0 +1,21 @@
...
+
+    return {
+        storeDesigner: (key: string, designer: CircuitDesigner) => {
+            storedDesigners.set(key, designer);
+        },
+        useDesigner:     (key: string) => storedDesigners.get(key),
+        useMainDesigner: () => {
+            const designer = storedDesigners.get("main");
+            if (!designer)
+                throw new Error("useMainCircuit: Failed to find a circuit labeled as 'main'!");
```
KevinMackenzie, line 15:
```suggestion
                throw new Error("useMainDesigner: Failed to find a circuit designer labeled as 'main'!");
```

src/site/shared/utils/SnapUtils.ts
```diff
@@ -0,0 +1,40 @@
+import {Component} from "core/public";
+import {V, Vector} from "Vector";
+
+
+const WIRE_SNAP_THRESHOLD = 0.2;
+
+/**
+ * Utility used for calculating positions of ports and wires when a component
+ * gets moved.
```
KevinMackenzie, line 9: Can you describe the behavior more specifically.  i.e. "Adjusts a proposed new position for a component to snap it into positions that result in axis-aligned wire connections."

KevinMackenzie (changes requested):
Disclaimer: i haven't run it at all.  Looks good overall, behavior makes sense, etc.  Most of the comments are on convention.

src/app/core/internal/view/rendering/prims/SVGPrim.ts
```diff
@@ -41,7 +42,7 @@ export class SVGPrim implements Prim {
         throw new Error("Method not implemented.");
     }
     public hitTest(pt: Vector): boolean {
-        throw new Error("Method not implemented.");
+        return RectContains(this.transform, pt);
```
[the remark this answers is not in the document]
LeonMontealegre: Yeah this is current behavior, but yeah maybe in the future we can have a more accurate hit-test

src/app/core/public/api/impl/Circuit.ts
```diff
@@ -118,19 +115,27 @@ export abstract class CircuitImpl<
     }
 
     // Queries
-    public pickObjectAt(pt: Vector): ComponentT | WireT | PortT | undefined {
-        throw new Error("Unimplemented");
+    public pickObjectAt(pt: Vector, space: Vector.Spaces = "world"): ComponentT | WireT | PortT | undefined {
+        const pos = ((space === "world") ? pt : this.view.toWorldPos(pt));
+        const objID = this.view.findNearestObj(pos);
```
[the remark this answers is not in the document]
LeonMontealegre: Hmmm, idk, this is an important decision though -- whether or not we want to expose `Option` to the API users

Will add to discussion notes

src/app/core/public/api/impl/Circuit.ts
```diff
@@ -118,19 +115,27 @@ export abstract class CircuitImpl<
...
     }
 
-    public selectedObjs(): Obj[] {
+    public get selectedObjs(): Obj[] {
         return this.selections.get()
                .map((id) => this.getObj(id))
                .filter((obj) => (obj !== undefined)) as Obj[];
     }
 
+    public hasObj(obj: Obj): boolean {
```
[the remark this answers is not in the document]
LeonMontealegre: ... this isn't even used anywhere I have no idea what this is lol

src/app/core/public/api/impl/Port.ts
```diff
@@ -34,6 +35,24 @@ export class PortImpl<
         return this.getObj().index;
     }
 
+    public get originPos(): Vector {
+        // TODO[model_refactor_api]: This probably needs to be calculated explicitly here ?
```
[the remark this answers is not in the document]
LeonMontealegre: Yeah, what if the assembler hasn't updated it yet after a change, or if it was just created?
Especially since (i think?) we wanted to move the assembling to happen at the beginning of the draw call or something so that if things were changed multiple times in succession we don't re-assemble multiple times needlessly?
I'll add to discussion notes too

src/site/shared/circuitdesigner/impl/CircuitDesigner.ts
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    public get cursor() {
+        return this.state.cursor;
+    }
+    public set cursor(cursor: Cursor | undefined) {
+        this.state.cursor = cursor;
+    }
+
+    public attachCanvas(canvas: HTMLCanvasElement): () => void {
```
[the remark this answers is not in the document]
LeonMontealegre: It makes it so that in react with a `useEffect` we can `return designer.attachCanvas()` since the effects need to return the [cleanup function](https://react.dev/reference/react/useEffect#parameters) -- this is a **very** useful thing to have imo (and also will be described in the documentation eventually)

src/site/shared/utils/SnapUtils.ts
LeonMontealegre: Yeah.. this is legacy code and this comment was probably written as RCOS fodder

src/site/shared/circuitdesigner/impl/CircuitDesigner.ts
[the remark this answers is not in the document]
KevinMackenzie: I'm not really convinced eliding the cleanup function from the `useEffect` call this way is worth it, sacrificing clarity and separation of concern somewhat for lower syntactic overhead.  I'm willing to hold my judgement for now if you alias this return type so the semantics are clear, i.e. `ReactCleanupFunction`, emphasis on `React`.  The idea being to at some point revisit whether this is a pattern that makes sense to use at this level.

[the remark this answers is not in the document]
LeonMontealegre: I disagree, it's simply a "cleanup" function, feels like a not-super-crazy pattern ?
Otherwise, it would just return `void` anyways so might as well return something useful that the user doesn't necessarily have to use but can be convenient?

src/app/core/internal/view/CircuitView.ts
```diff
@@ -103,6 +105,19 @@ export abstract class CircuitView {
         return this.cameraMat.inverse().mul(pos).add(this.renderer.size.scale(0.5));
     }
 
+    public findNearestObj(pos: Vector): GUID | undefined {
+        for (const [id, prims] of this.componentPrims) {
+            if (prims.some((prim) => prim.hitTest(pos)))
+                return id;
+            // TODO: hit test the component's ports as well
```
TGCrystal, line 112: TODO syntax

src/app/core/public/api/impl/Circuit.ts
```diff
@@ -118,14 +115,18 @@ export abstract class CircuitImpl<
...
 
     // Queries
-    public pickObjectAt(pt: Vector): ComponentT | WireT | PortT | undefined {
-        throw new Error("Unimplemented");
+    public pickObjectAt(pt: Vector, space: Vector.Spaces = "world"): ComponentT | WireT | PortT | undefined {
+        const pos = ((space === "world") ? pt : this.view.toWorldPos(pt));
+        const objID = this.view.findNearestObj(pos);
+        if (!objID)
+            return undefined;
+        return this.getObj(objID);
```
TGCrystal, line 123: Could be
```typescript
return !objID ? undefined : this.getObj(objID);
```
Can't remember whether you liked ternary for things like this

src/site/shared/tools/DefaultTool.ts
```diff
@@ -10,30 +10,27 @@ export class DefaultTool {
...
-        //         ((o) => (o.baseKind !== "Port"))
-        //     )?.id;
-        // }
-        // if (ev.type === "mouseup")
-        //     info.curPressedObjID = undefined;
+        if (ev.type === "mousedown") {
+            // Find object if we pressed on one
+            designer.curPressedObj = designer.circuit.pickObjectAt(ev.state.mousePos, "screen");
+        }
+        if (ev.type === "mouseup")
```
TGCrystal, line 27: Can this be an `else if`?

src/site/shared/tools/TranslateTool.ts
```diff
@@ -0,0 +1,71 @@
...
+        // If the pressed component is part of the selected objects,
+        //  then translate all of the selected objects
+        //  otherwise, just translate the pressed object
+        this.components = (!curPressedObj || curPressedObj.isSelected)
+            ? (circuit.selectedObjs.filter((o) => (o.baseKind === "Component"))) as Component[]
+            : [curPressedObj as Component];
+
+        circuit.beginTransaction();
+
+        // TODO: shift components
```
TGCrystal, line 41: TODO syntax

src/app/core/public/api/impl/Port.ts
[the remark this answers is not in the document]
LeonMontealegre: We should `dirty` the prims that are changed, and then when we need information from them, we can re-compute at that point

src/app/core/public/api/impl/Circuit.ts
LeonMontealegre: eh yeah not really in this case, I think undefined checks should remain if-statements for the most part

KevinMackenzie (approved):

PR #1249, opened 2023-04-06
[3] Added wiring tool and "tool renderers"

LeonMontealegre:
Added infrastructure for "tool renderers" that will draw on-top of the canvas for tool-things, i.e. a fake wire for WiringTool.

### DIFF
--- a/src/app/core/public/api/Circuit.ts
+++ b/src/app/core/public/api/Circuit.ts
@@ -11,6 +11,8 @@ import {ComponentInfo} from "./ComponentInfo";
 import {Obj}           from "./Obj";
 import {Port}          from "./Port";
 import {Wire}          from "./Wire";
+import {RenderHelper}  from "core/internal/view/rendering/RenderHelper";
+import {RenderOptions} from "core/internal/view/rendering/RenderOptions";
 
 
 export type {CircuitMetadata} from "core/schema/CircuitMetadata";
@@ -34,8 +36,11 @@ export interface Circuit {
     readonly camera: Camera;
 
     // Queries
-    pickObjectAt(pt: Vector, space?: Vector.Spaces): Obj | undefined;
-    pickObjectRange(bounds: Rect): Obj[];
+    pickObjAt(pt: Vector, space?: Vector.Spaces): Obj | undefined;
+    pickComponentAt(pt: Vector, space?: Vector.Spaces): Component | undefined;
+    pickWireAt(pt: Vector, space?: Vector.Spaces): Wire | undefined;
+    pickPortAt(pt: Vector, space?: Vector.Spaces): Port | undefined;
+    pickObjRange(bounds: Rect): Obj[];
     readonly selectedObjs: Obj[];
 
     getComponent(id: GUID): Component | undefined;
@@ -82,7 +87,10 @@ export interface Circuit {
     attachCanvas(canvas: HTMLCanvasElement): () => void;
     detachCanvas(): void;
 
-    addRenderCallback(cb: () => void): void;
+    forceRedraw(): void;
+
+    // TODO[](leon) - Need to make a public-facing RenderHelper/RenderOptions
+    addRenderCallback(cb: (data: { renderer: RenderHelper, options: RenderOptions, circuit: Circuit }) => void): void;
 
     subscribe(cb: (ev: any) => void): () => void;
 }
--- a/src/app/core/public/api/impl/Circuit.ts
+++ b/src/app/core/public/api/impl/Circuit.ts
@@ -20,6 +20,8 @@ import {CircuitState}         from "./CircuitState";
 import {CreateDrawingFromSVG} from "svg2canvas";
 import {CameraImpl}           from "./Camera";
 import {Observable}           from "core/utils/Observable";
+import {RenderHelper}         from "core/internal/view/rendering/RenderHelper";
+import {RenderOptions}        from "core/internal/view/rendering/RenderOptions";
 
 
 export abstract class CircuitImpl<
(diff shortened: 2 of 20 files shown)


TGCrystal (dismissed):

src/app/core/utils/Reducers.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Utility reducer that can reduce a generic object with a specific "dist"
+ *  parameter. This finding an object with the minimum "distance" and then
```
KevinMackenzie, line 3: This second sentence reads weird

src/site/shared/tools/renderers/ToolRenderer.ts
```diff
@@ -0,0 +1,24 @@
+import {RenderHelper}  from "core/internal/view/rendering/RenderHelper";
+import {RenderOptions} from "core/internal/view/rendering/RenderOptions";
+import {Circuit}       from "core/public";
+
+import {InputManagerState} from "shared/utils/input/InputManagerState";
+
+import {Tool} from "../Tool";
+
+
+export interface ToolRendererProps<T extends Tool = Tool> {
```
KevinMackenzie, line 10: Thoughts on `ToolRendererArgs`?

src/site/shared/circuitdesigner/impl/CircuitDesigner.ts
```diff
@@ -39,6 +43,17 @@ export class CircuitDesignerImpl<CircuitT extends Circuit> implements CircuitDes
 
         // Setup tool manager
         this.inputAdapter.subscribe((ev) => this.toolManager.onEvent(ev, this));
+
+        // Attach tool renderers
+        if (renderers) {
+            this.circuit.addRenderCallback(({ renderer, options, circuit }) => {
+                renderers.forEach((toolRenderer) => {
+                    const curTool = this.toolManager.curTool;
+                    if (toolRenderer.isActive(curTool))
```
KevinMackenzie, line 52: Could you keep track of which tool renderer(s) are associated with each tool on construction and avoid this`isActive` pattern.  Making it something like:

```typescript
this.circuit.addRenderCallback(({ renderer, options, circuit }) => {
    this.toolManager.curToolRenderer.render({ renderer, options, circuit, this.toolManager.curTool, input: this.inputAdapter.state });
});
```
And then encapsulate each tool renderer with its associated tool once on construction for static type safety.

src/site/shared/circuitdesigner/index.ts
```diff
@@ -1,16 +1,10 @@
...
+import {CircuitDesignerImpl, ToolConfig} from "./impl/CircuitDesigner";
 
 
 export * from "./CircuitDesigner";
 
-export function CreateDesigner<Circ extends Circuit>(circuit: Circ, toolConfig: {
-    defaultTool: DefaultTool;
-    tools: Tool[];
-}) {
+export function CreateDesigner<Circ extends Circuit>(circuit: Circ, toolConfig: ToolConfig) {
```
KevinMackenzie, line 8: Could you make it `CircuitT extends Circuit` like the comment on the [2] PR?

src/site/shared/tools/WiringTool.ts
```diff
@@ -0,0 +1,109 @@
...
+import {CircuitDesigner}                       from "shared/circuitdesigner/CircuitDesigner";
+import {LEFT_MOUSE_BUTTON, RIGHT_MOUSE_BUTTON} from "shared/utils/input/Constants";
+import {InputManagerEvent}                     from "shared/utils/input/InputManagerEvent";
+import {Tool}                                  from "./Tool";
+
+
+// The distance away from the port in selection for wiring.
+export const WIRING_PORT_SELECT_RADIUS = 0.34;
+
+export class WiringTool implements Tool {
```
KevinMackenzie, line 18: Is this all a direct port of the previous impl?

src/app/core/public/api/impl/Circuit.ts
```diff
@@ -113,11 +115,26 @@ export abstract class CircuitImpl<
     }
 
     // Queries
-    public pickObjectAt(pt: Vector, space: Vector.Spaces = "world"): ComponentT | WireT | PortT | undefined {
+    public pickObjAt(pt: Vector, space: Vector.Spaces = "world"): ComponentT | WireT | PortT | undefined {
         const pos = ((space === "world") ? pt : this.view.toWorldPos(pt));
         return this.view.findNearestObj(pos).map((id) => this.getObj(id)).asUnion();
     }
-    public pickObjectRange(bounds: Rect): Array<ComponentT | WireT | PortT> {
+    public pickComponentAt(pt: Vector, space: Vector.Spaces = "world"): ComponentT | undefined {
```
KevinMackenzie, line 122: Could you try to factor out the common structure from these 3 functions (and maybe even the `pickObjAt`)?  NBD if that shows up in a later PR in this chain of PRs.

KevinMackenzie (changes requested):
A couple of small requests and one medium-sized question about storing `ToolRenderer`s

src/site/shared/tools/renderers/ToolRenderer.ts
LeonMontealegre: I like

src/app/core/utils/Reducers.ts
LeonMontealegre: Jesus I don't even know what I was trying to write there good lord

src/site/shared/tools/WiringTool.ts
LeonMontealegre: Mostly yes, cleaned up parts where I thought I could -- please feel free to comment on its implementation anyways since it's good to talk about

src/site/shared/circuitdesigner/impl/CircuitDesigner.ts
LeonMontealegre: I don't think so -- like the Rotate Tool (as seen in future PRs) will need to be active while rotating but also for the DefaultTool when something is selected to show the rotation circle, and theoretically other tool renderers could also effect the DefaultTool

src/app/core/public/api/impl/Circuit.ts
LeonMontealegre: I attempted -- can you take another look?

src/app/core/public/api/Circuit.ts
```diff
@@ -82,7 +87,10 @@ export interface Circuit {
     attachCanvas(canvas: HTMLCanvasElement): () => void;
     detachCanvas(): void;
 
-    addRenderCallback(cb: () => void): void;
+    forceRedraw(): void;
+
+    // TODO[](leon) - Need to make a public-facing RenderHelper/RenderOptions
+    addRenderCallback(cb: (data: { renderer: RenderHelper, options: RenderOptions, circuit: Circuit }) => void): void;
```
LeonMontealegre, line 93: Also what do you think about this @KevinMackenzie ?

KevinMackenzie: It's a little gross as-is, but it seems like a reasonable compromise until we separate the existing `CircuitView` from the rendering itself, as mentioned in the IC's discussion.  At that point the render callback would exist on the "Viewport" type rather than the Circuit type and most/all of the `RenderHelper`/`RenderOptions` would live outside `API.Circuit` too.

src/site/shared/circuitdesigner/impl/CircuitDesigner.ts
KevinMackenzie: I see.  I think this is a limitation of the current Tools state machine model, then.  A model that might fit this Rotate Tool use-case could be: 
- Each tool has one state of; `Active`, `Pending`, and `Inactive`.
- At any given time there is one `Active` tool.
- All tools that are accessible from the current state are `Pending`.
- All other tools are `Inactive`

Each tool renderer is run every frame, deciding what, if anything, should be rendered based on the state of its associated tool.  For example, when a component is selected, the rotate tool is `Pending`, and so the rotate tool renderer would render the tool for the `Pending` state.

This is an over-simplification, but do you understand what I'm getting at?

src/app/core/public/api/impl/Circuit.ts
KevinMackenzie: Definitely better!

src/site/shared/tools/WiringTool.ts
KevinMackenzie: Ok, yea, I'm not super tuned-in to the tool state flow, but `findPort` makes sense at least

src/site/shared/circuitdesigner/impl/CircuitDesigner.ts
LeonMontealegre: Going to handle this in a following PR as not to clutter this one

KevinMackenzie (approved):

PR #1250, opened 2023-04-06
[4] Added rotate tool and selection box tool

LeonMontealegre:
Most of their implementations are copied from how they have been

Changed how InputAdapter works, so that instead of constructing it once and then setting up/tearing down on it, instead construct it to set it up and then tear it down when necessary and re-construct it to set it up again. Removes need to `!` every access of `canvas` in the InputAdapter, and also removes need to have an optional instance of it in `CircuitDesigner`, instead it's just held locally. 
This does come at the cost of not having an explicit `tearDown` method anymore on `CircuitDesigner` and instead needing to use the callback `tearDown`.

### DIFF
--- a/src/app/core/internal/view/rendering/prims/BaseShapePrim.ts
+++ b/src/app/core/internal/view/rendering/prims/BaseShapePrim.ts
@@ -31,6 +31,8 @@ export abstract class BaseShapePrim implements Prim {
             ctx.strokeStyle = this.style.stroke.color;
             ctx.lineWidth = this.style.stroke.size;
         }
+        if (this.style.alpha !== undefined)
+            ctx.globalAlpha = this.style.alpha;
 
         ctx.save();
 
@@ -48,6 +50,10 @@ export abstract class BaseShapePrim implements Prim {
         ctx.closePath();
 
         ctx.restore();
+
+        // Reset alpha if we set it
+        if (this.style.alpha !== undefined)
+            ctx.globalAlpha = 1;
     }
 
     public updateStyle(style: Style): void {
--- a/src/site/shared/tools/renderers/RotateToolRenderer.ts
+++ b/src/site/shared/tools/renderers/RotateToolRenderer.ts
@@ -1,0 +1,45 @@
+import {CirclePrim}       from "core/internal/view/rendering/prims/CirclePrim";
+import {CircleSectorPrim} from "core/internal/view/rendering/prims/CircleSectorPrim";
+
+import {ROTATION_CIRCLE_RADIUS, ROTATION_CIRCLE_THICKNESS, RotateTool} from "../RotateTool";
+
+import {ToolRenderer} from "./ToolRenderer";
+
+
+export const RotateToolRenderer: ToolRenderer<RotateTool | undefined> = {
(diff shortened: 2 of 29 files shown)


src/app/core/internal/view/rendering/prims/BaseShapePrim.ts
```diff
@@ -48,6 +50,10 @@ export abstract class BaseShapePrim implements Prim {
         ctx.closePath();
 
         ctx.restore();
+
+        // Reset alpha if we set it
+        if (this.style.alpha !== undefined)
```
KevinMackenzie, line 55: is this `!==` required because of `number`'s truthiness?

src/site/shared/tools/renderers/RotateToolRenderer.ts
```diff
@@ -0,0 +1,45 @@
...
+            }));
+        }
+
+        const selections = circuit.selectedObjs;
+
+        // If we are in the default tool, draw the rotation circle outline if we have only components selected
+        if (!curTool) {
+            if (selections.length > 0 &&
+                selections.every((o) => (o.baseKind === "Component"))) {
+                drawOutline();
```
KevinMackenzie, line 31: This kind of indentation is kind of hard to read.  Usually I see this mitigated by putting the logical operators at the start of the line.  What's the policy on this?

src/site/shared/utils/input/InputAdapter.ts
```diff
@@ -72,53 +81,39 @@ export class InputAdapter extends Observable<InputAdapterEvent> {
     private readonly dragTime: number;
 
     /** The canvas the user is performing inputs on. */
-    private canvas?: HTMLCanvasElement;
+    private readonly canvas: HTMLCanvasElement;
```
KevinMackenzie, line 84: Nice

KevinMackenzie (changes requested):
Just one minor thing.

src/app/core/internal/view/rendering/prims/BaseShapePrim.ts
LeonMontealegre: Yep  :D

src/site/shared/tools/renderers/RotateToolRenderer.ts
LeonMontealegre: Yeah... I don't have a great solution here, I think I've been pretty consistent with having the logical operators at the ends of lines -- I think best policy may be to keep complex expressions out of if-statements generally anyways ? not sure

KevinMackenzie: > ...I think best policy may be to keep complex expressions out of if-statements generally anyways?...
I think this is a tough sell as a blanket policy, but I don't have a good solution either.

KevinMackenzie: It's kind of avoiding the problem, but i'd maybe factor out the lambda into a helper function somewhere so 29/30 fit on one line.  I imagine we want this kind of "base kind" filtering somewhere else too anyway ( :

TGCrystal (dismissed):

KevinMackenzie (approved):

TGCrystal (approved):

PR #1251, opened 2023-04-06
[5] Added split wire tool, node, and split wire methods

LeonMontealegre:


src/app/core/public/api/impl/Circuit.ts
```diff
@@ -246,8 +246,9 @@ export abstract class CircuitImpl<
         throw new Error("Unimplemented");
     }
     public clearSelections(): void {
-        // TODO(callac5)
-        throw new Error("Unimplemented");
+        this.beginTransaction();
+        this.selections.clear();
+        this.commitTransaction();
```
KevinMackenzie, line 251: I don't think this needs to be in a transaction

src/app/core/public/api/impl/Wire.ts
```diff
@@ -24,10 +25,37 @@ export class WireImpl<
             .unwrap();
     }
 
+    protected abstract get nodeKind(): string;
```
KevinMackenzie, line 28: assuming this is temporary based on TODO below, and will be moved somewhere else as a part of that?  (i.e. `ObjectInfo.nodeKind(): string` or eventually something more general that includes busses.

src/app/core/public/api/Port.ts
```diff
@@ -29,5 +29,5 @@ export interface Port extends BaseObject {
     readonly connectedPorts: Port[]; // TODO: think of a better name for this ?
 
     getLegalWires(): Port.LegalWiresQuery;
-    connectTo(other: Port): void;
+    connectTo(other: Port | Component): Wire | undefined;
```
KevinMackenzie, line 32: I have mixed feelings about this taking `Port | Component`

KevinMackenzie (dismissed):

TGCrystal (dismissed):

TGCrystal (approved):

KevinMackenzie (approved):

PR #1254, opened 2023-04-17
Re-add all Tool Handlers

LeonMontealegre:
Also add `Node` API type.

Most handlers are still non-functional, but they all depend on API methods that need to be implemented.

### DIFF
--- a/src/app/core/public/api/impl/Selections.ts
+++ b/src/app/core/public/api/impl/Selections.ts
@@ -27,14 +27,17 @@ export class SelectionsImpl implements Selections {
     }
 
     public get all(): Obj[] {
-        return this.selections.map((id) => this.state.getObj(id)!);
+        return this.selections
+            .map((id) => this.state.getObj(id)!);
     }
     public get components(): Component[] {
-        return this.selections.filter((id) => (this.doc.hasComp(id)))
+        return this.selections
+            .filter((id) => this.doc.hasComp(id))
             .map((id) => this.state.constructComponent(id));
     }
     public get wires(): Wire[] {
-        return this.selections.filter((id) => (this.doc.hasWire(id)))
+        return this.selections
+            .filter((id) => this.doc.hasWire(id))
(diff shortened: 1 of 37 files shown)


src/app/core/public/api/impl/Selections.ts
```diff
@@ -22,21 +23,39 @@ export class SelectionsImpl implements Selections {
...
+        return this.selections.map((id) => this.state.getObj(id)!);
+    }
     public get components(): Component[] {
         return this.selections.filter((id) => (this.doc.hasComp(id)))
             .map((id) => this.state.constructComponent(id));
     }
+    public get wires(): Wire[] {
+        return this.selections.filter((id) => (this.doc.hasWire(id)))
+            .map((id) => this.state.constructWire(id));
+    }
```
TGCrystal, line 39: I think this would look better with the `.filter` on its own line like the `forEach` function below

src/site/shared/circuitdesigner/impl/CircuitDesigner.ts
```diff
@@ -64,16 +64,12 @@ export class CircuitDesignerImpl<CircuitT extends Circuit> implements CircuitDes
...
-                        toolRenderer.render({ renderer, options, circuit, curTool, input: inputAdapter.state });
-                });
-            });
-        }
+        const renderCleanup = this.circuit.addRenderCallback((renderArgs) =>
+            renderers.forEach((toolRenderer) => toolRenderer.render({
+                curTool: this.toolManager.curTool,
+                input:   inputAdapter.state,
+                ...renderArgs,
+             })));
```
TGCrystal, line 72: Can the last `));` be moved to the next line so it is a little clearer what they are closing?

TGCrystal (changes requested):

TGCrystal (approved):

KevinMackenzie (approved):
No major issues (or any that I remember)

PR #1274, opened 2023-04-24
Fix Headless Circuits

LeonMontealegre:
Currently, the Circuit expects for a view, canvas, and images to exist. We want the Circuit to be able to exist in a "headless" state where the view can exists virtually and no rendering needs to be performed, but all assembled Prims can still be accessed for computed properties such as port positions.

This also allows images/assets to be lazy-loaded and fixes all of the units test failings.

### DIFF
--- a/src/app/core/internal/view/CircuitView.ts
+++ b/src/app/core/internal/view/CircuitView.ts
@@ -23,6 +23,13 @@ import {DefaultRenderOptions, RenderOptions} from "./rendering/RenderOptions";
 import {RenderScheduler}                     from "./rendering/RenderScheduler";
 
 
+/**
+ * Utility class to manage assets for the circuit view.
+ *
+ * Specifically used over a Map so that it can be observed so that when an
+ * asset is updated (i.e. set for the first time), dependencies of that
+ * assets can be notified and update accordingly.
+ */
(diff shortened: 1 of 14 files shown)


src/app/core/internal/view/CircuitView.ts
```diff
@@ -67,44 +101,52 @@ export abstract class CircuitView extends Observable<{ renderer: RenderHelper }>
...
+                    if (props.has("x") || props.has("y") || props.has("angle")) {
+                        const ports = this.circuit.doc.getPortsForComponent(id);
+                        ports.map((ports) => ports.forEach((portID) => {
+                            this.circuit.doc.getWiresForPort(portID)
+                                .map((wires) => wires.forEach((wireID) => this.dirtyWires.add(wireID)))
+                        }))
+                    }
+                } else if (circuit.doc.hasWire(id))
+                    this.dirtyWires.add(id);
+                else if (circuit.doc.hasPort(id))
```
TGCrystal, line 148: If the `if` has `{}`, then the `else if`s probably should as well

TGCrystal (changes requested):
Merged this into my expression parser branch and the test I was having issues with worked! Just the one little formatting thing

KevinMackenzie (approved):

PR #1278, opened 2023-05-07
Add Circuit.find and findAll methods to replace other get methods

LeonMontealegre:
I'm very much still unsure if this change is worth making, my solution feels hacky but unfortunately it's the only way I found to make it work.

The problem resides in that apparently templated methods like the `find` method cannot be acted on by something like `ToDigital` as it will strip it of it's templating. I think this is related to something called ["higher kinded types"](https://github.com/Microsoft/TypeScript/issues/1213), which is just a fundamental TS limitation.

That means each sub-project of the API will have to declare the find/findAll methods explicitly with their types which might not be that big of a deal and we can further make some core utility for typing it explicitly but it's not ... great

If these end up being the only instances of templated methods then it might just simply not be worth the effort but idk.

Not sure it even really cleans up the usage much (at least currently), but theoretically it could allow for some nice fancy querying when the queries are more complex than "get all components".

Thoughts?

### DIFF
--- a/src/app/core/internal/view/CircuitView.ts
+++ b/src/app/core/internal/view/CircuitView.ts
@@ -67,6 +67,7 @@ export abstract class CircuitView extends Observable<{ renderer: RenderHelper }>
 
     public localPortPositions: Map<GUID, PortPos>; // Key'd by port ID
     public portPositions: Map<GUID, PortPos>; // Key'd by port ID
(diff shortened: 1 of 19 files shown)


TGCrystal:
I feel like we had examples where this would be nice but I can't remember them now. Can you remember any more advanced use cases that make this worthwhile?

PR #1284, opened 2023-12-21
Model refactor API SelectionPopup [1]

LeonMontealegre:
Added back initial parts of SelectionPopup
No modules yet

### DIFF
--- a/src/app/core/internal/impl/CircuitInternal.ts
+++ b/src/app/core/internal/impl/CircuitInternal.ts
@@ -12,6 +12,19 @@ import {CircuitDocument, ReadonlyCircuitDocument}        from "./CircuitDocument
 import {FastCircuitDiff, FastCircuitDiffBuilder}         from "./FastCircuitDiff";
 
 
+export type InternalEvent = {
+    type: "CircuitOp";
+    diff: FastCircuitDiff;
+} | {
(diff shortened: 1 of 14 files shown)


src/app/core/internal/impl/CircuitInternal.ts
```diff
@@ -303,12 +316,22 @@ export class CircuitInternal extends Observable<FastCircuitDiff> {
     }
 
     public setCameraProps(props: Partial<Schema.Camera>) {
+        const dx = (props.x ?? this.camera.x) - this.camera.x;
+        const dy = (props.y ?? this.camera.y) - this.camera.y;
+        const dz = (props.zoom ?? this.camera.zoom) - this.camera.zoom;
+
+        // No change, no need to emit event
+        if (dx === 0 && dy === 0 && dz === 0)
```
TGCrystal, line 324: iirc the positions can be floats. Do you want to instead check that each is below some small value? Or does the whole usage of the null coalescing operator ensure that won't happen?

TGCrystal (approved):

src/app/core/internal/impl/CircuitInternal.ts
LeonMontealegre: the latter, yeah

PR #1288, opened 2024-01-29
Model refactor API Popups

LeonMontealegre:
Added initial popups back to the site

### DIFF
--- a/src/site/shared/containers/ImageExporterPopup/index.scss
+++ b/src/site/shared/containers/ImageExporterPopup/index.scss
@@ -163,3 +163,28 @@
         }
     }
 }
+
+.image-exporter-preview__button {
+    position: absolute;
+
+    top: 15px;
+    right: 15px;
+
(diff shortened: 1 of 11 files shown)


src/site/shared/containers/ImageExporterPopup/index.scss
```diff
@@ -163,3 +163,28 @@
         }
     }
 }
+
+.image-exporter-preview__button {
```
TGCrystal, line 167: Did we have some sort of style convention that dictated both the mix of `-` and `_` and also the double `_` between `preview` and `button`?

TGCrystal (approved):

src/site/shared/containers/ImageExporterPopup/index.scss
LeonMontealegre: this was just from before, we can have stylelint do naming stuff later

PR #1295, opened 2025-02-09
Add back in OR and XOR gate rendering

TGCrystal:


src/app/digital/internal/assembly/components/ORGateAssembler.ts
```diff
@@ -1,143 +1,142 @@
...
+                return;
+            }
 
-        line.forEach((q) => image.push(q));
-        image.push(img);
+            const selected = this.selections.has(gate.id);
 
-        this.view.componentPrims.set(gate.id, image);
+            quadCurves.forEach((quadCurve) => {
+                // TODO: better type safe way of handling this?
```
LeonMontealegre, line 136: Yeah... we need a much better system for this, I hate it so much lol; but it's the correct way for now

LeonMontealegre (approved):

PR #1308, opened 2025-03-02
ConstantNumberAssembly and fix Constant High/Low

TGCrystal:


src/projects/digital/api/circuit/src/internal/assembly/components/ConstantHighAssembler.ts
```diff
@@ -13,7 +13,10 @@ export class ConstantHighAssembler extends ComponentAssembler {
 
     public constructor(params: AssemblerParams, sim: DigitalSim) {
         super(params, V(1, 1), {
-            "outputs": () => ({origin: V(0.5, 0), dir: V(1, 0)})
+            "outputs": () => ({
+                origin: V(this.size.x / 2 - this.options.defaultBorderWidth, 0),
```
LeonMontealegre, line 17: You can at least hard-code the `0.5` probably (instead of `this.size.x / 2` -- but can you also add a comment explaining why this is necessary?

src/projects/shared/api/circuit/src/internal/assembly/PrimBounds.ts
```diff
@@ -22,6 +22,9 @@ export function Bounds(prim: Prim): Rect {
         return prim.transform.asRect();
     case "SVG":
         return prim.transform.asRect();
+    case "Text":
+        // Text doesn't know its own size because that requires canvas
```
LeonMontealegre, line 26: Add to this comment saying something like "so make sure you have another prim surrounding it other it will get culled" or something like that

src/projects/shared/api/circuit/src/internal/assembly/Prim.ts
```diff
@@ -70,14 +70,26 @@ export interface RectanglePrim {
...
 export type BaseShapePrim =
     | BezierCurvePrim
     | CirclePrim
     | CircleSectorPrim
     | LinePrim
     | PolygonPrim
     | QuadCurvePrim
-    | RectanglePrim;
+    | RectanglePrim
+    | TextPrim;
```
LeonMontealegre, line 92: I don't think TextPrim should be a "BaseShape", kinda semantically but also I think it doesn't need a "Style" attribute and that's what the "font" should cover (you can also add a attribute for color I guess)

LeonMontealegre (changes requested):

src/projects/digital/api/circuit/src/internal/assembly/components/ConstantNumberAssembler.ts
```diff
@@ -0,0 +1,118 @@
...
+                kind: "Text",
+
+                dependencies: new Set([AssemblyReason.TransformChanged, AssemblyReason.SelectionChanged, AssemblyReason.PropChanged]),
+                assemble:     (comp) => this.assembleText(comp),
+
+                getFontStyle: () => this.getFontStyle(),
+            },
+        ]);
+        this.sim = sim;
+        this.info = this.circuit.getComponentInfo("Switch").unwrap() as DigitalComponentInfo;
```
LeonMontealegre, line 48: Definitely should not be using the Switch's component info

src/projects/digital/api/circuit/src/internal/assembly/components/ConstantNumberAssembler.ts
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    private getOutputLocations(index: number) {
+        const targetY = .75 - (index * 1.5 / 3);
+        return { origin: V((this.size.x - this.options.defaultBorderWidth) / 2, targetY), target: V(1.2, targetY) } as const
+    }
+
+    private assembleRectangle(comp: Schema.Component) {
+        const transform = this.getTransform(comp).copy();
+        transform.setSize(transform.getSize().sub(V(this.options.defaultBorderWidth)));
```
LeonMontealegre, line 58: nit: mutability is gross, just do `const transform = new Transform(this.getPos(comp), this.getAngle(comp), this.getSize(comp).sub(V(this.options.defaultBorderWidth))` --- but actually, why do you need to subtract the border width? in the constant low/high you changed the port's origin instead? (add a comment if you think it's still necessary)

src/projects/digital/api/circuit/src/internal/assembly/components/ConstantNumberAssembler.ts
```diff
@@ -0,0 +1,118 @@
...
+            p2: transform.toWorldSpace(V(x, y2)),
+        } as const;
+    }
+
+    private getLineStyle(comp: Schema.Component): Style {
+        const style = this.options.lineStyle(this.isSelected(comp.id));
+        const { stroke } = style
+        return {
+            ...style,
+            stroke: stroke? { ...stroke, lineCap: "square" } : undefined,
```
LeonMontealegre, line 86: nit: space before `?`

LeonMontealegre (approved):

src/projects/digital/api/circuit/src/internal/assembly/components/ConstantNumberAssembler.ts
TGCrystal: This isn't really about getting ports to line up. Not including the border, the constant number should be the same size as constant high/low.
nightly:
image
without subtraction:
image
with subtraction:
image

package.json
```diff
@@ -113,5 +113,9 @@
       "last 1 firefox version",
       "last 1 safari version"
     ]
+  },
+  "volta": {
```
TGCrystal, line 117: Added this because I thought node was bugging out on me. It wasn't, I just was putting `ConstantNumberAssembler` into getComponentInfo instead of `ConstantNumber`. Left it in though because Volta is a tool that automatically swaps to node/yarn versions you have defined in package.json like this, so nice to have. And it'll make future upgrades of these easier

LeonMontealegre (approved):

PR #1315, opened 2025-03-10
Model refactor api smart place

TGCrystal:


src/projects/shared/api/circuit/src/utils/Array.ts
```diff
@@ -14,7 +14,7 @@ declare global {
 }
 
 Array.prototype.sum = function(this: number[]): number {
-    return this.reduce((a, b) => (a + b));
+    return this.length === 0 ? 0 : this.reduce((a, b) => (a + b));
```
LeonMontealegre, line 17: nit: just do `this.reduce((a, b) => (a + b), 0);` instead

src/projects/digital/site/src/containers/DigitalMainDesigner/index.tsx
```diff
@@ -8,11 +8,70 @@ import {MainDesigner} from "shared/site/containers/MainDesigner";
...
+ * @param options The options used to indicate what connected components to create.
+ * @throws If the itemId is an invalid item or IC.
+ */
+export function SmartPlace(pos: Vector, itemId: string, circuit: DigitalCircuit,
+    N: number, options: SmartPlaceOptions) {
+    circuit.beginTransaction();
+    for (let i = 0; i < N; i++) {
+        const comp = circuit.placeComponentAt(itemId, pos);
+
+        // Need to do it like this rather then comp.getInputPorts() since this can
```
LeonMontealegre, line 41: This is technically still accurate, `comp.inputs` would return the group labeled "inputs", not all the input ports (at the moment) -- maybe it would make more sense for it to return all the ports that are input ports but im not sure if that would be confusing or not, wdyt? This would also make a clear decision on the "undefined vs []" problem

LeonMontealegre (approved):

PR #1346, opened 2025-04-26
Fix migrated IC port positions

TGCrystal:


src/projects/digital/site/src/utils/DigitalVersionConflictResolver.ts
```diff
@@ -607,6 +607,9 @@ export function VersionConflictResolver(fileContents: string): VersionConflictRe
         const { guidsToObjs, objects, pinRefToPortGuid } = migrateObjs(objectRefsEntry, wiresRefsEntry, icGuids, true);
         const initialSimState = migrateSimState(guidsToObjs, icGuidsToObjects);
 
+        const displayWidth = x / 50;
+        const displayHeight = y / 50;
```
LeonMontealegre, line 611: Can you rename x and y to w and h or something, just do the / 50 down below when you set displayWidth, and then just do `x / (w / 2)` or something in the pin position? having to divide by 50 and then multiply by 25 below feels gross

PR #1347, opened 2025-04-26
Initial Proto implementation

LeonMontealegre:
- I would like to .gitignore the generated .ts files from .proto and automatically generate them with Webpack but was too lazy to do it here


TGCrystal (commented):
Changes look good, can you also add .proto files to .gitattributes? Would just look like `*.proto text`

TGCrystal (approved):

PR #1350, opened 2025-04-27
Added back (initial) IC Viewer

LeonMontealegre:
Currently missing https://github.com/OpenCircuits/OpenCircuits/issues/754


src/projects/shared/api/circuit/src/public/impl/Component.ts
```diff
@@ -115,6 +115,7 @@ export class ComponentImpl<T extends CircuitTypes> extends BaseObjectImpl<T> imp
     }
 
     public toSchema(): Schema.Component {
-        return ({ ...this.getComponent() });
+        const comp = this.getComponent();
+        return ({ ...comp, props: { ...comp.props } });
```
TGCrystal, line 119: I'm assuming this is because we don't want a shallow copy of props?

TGCrystal (approved):

src/projects/shared/api/circuit/src/public/impl/Component.ts
LeonMontealegre: yeah

PR #1355, opened 2025-05-01
Fix imported ICs not showing up in item nav

TGCrystal:
Also makes the overwrite warning when loading a circuit actually show up

### DIFF
--- a/src/projects/digital/site/src/containers/DigitalItemNav/index.tsx
+++ b/src/projects/digital/site/src/containers/DigitalItemNav/index.tsx
@@ -40,21 +40,27 @@ export const DigitalItemNav = () => {
         );
     });
 
+    const updateICs = () => setState({
+        ics: circuit.getICs().map((d) => ({
+            kind:      d.id,
+            label:     d.name,
+            icon:      "multiplexer.svg",
+            removable: true,
+        })),
+    });
+
     // Subscribe to CircuitDesigner
     //  and update the ItemNav w/
     //  ICs whenever they're added/removed
-    useEffect(() => circuit.subscribe((ev) => {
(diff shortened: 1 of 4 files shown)


src/projects/digital/site/src/containers/DigitalItemNav/index.tsx
```diff
@@ -40,21 +40,27 @@ export const DigitalItemNav = () => {
         );
     });
 
+    const updateICs = () => setState({
```
LeonMontealegre, line 43: put it inside the useEffect otherwise you have to add it as a dependency and wrap it in a useCallback and all the other react BS (might not have to necessarily with react compiler now but still i would be safe, especially since `circuit` CAN change now)

src/projects/shared/site/src/containers/Header/Right/OpenFileButton.tsx
```diff
@@ -19,7 +19,7 @@ export const OpenFileButton = () => {
     const fileInput = React.useRef<HTMLInputElement>(null);
 
     const load = (file: File) => {
-        if (isSaved || window.confirm(OVERWRITE_CIRCUIT_MESSAGE))
+        if (isSaved && window.confirm(OVERWRITE_CIRCUIT_MESSAGE))
```
LeonMontealegre, line 22: ? The logic is supposed to be "if the circuit is already saved" OR they confirm overwriting it, then load it. If it's an AND, then they have to save their circuit and confirm overwriting it; if it's not working it's probably an issue with isSaved not being set correctly?

LeonMontealegre (changes requested):

LeonMontealegre (approved):

PR #1399, opened 2025-05-11
Name transactions so the names show up in the history box

TGCrystal:
Example:
image


src/projects/digital/api/circuitdesigner/src/tools/ICPortTool.ts
```diff
@@ -78,7 +78,7 @@ export class ICPortTool extends ObservableImpl<ToolEvent> implements Tool {
     public onDeactivate(ev: InputAdapterEvent, { circuit }: CircuitDesigner): void {
         this.curPin = undefined;
 
-        circuit.commitTransaction();
+        circuit.commitTransaction("IC Port Tool");
```
LeonMontealegre, line 81: ik this is probably how master is doing it, but saying 'Tool' feels wrong; I think the rule of thumb should be entry names should be "past-tense-verbs on the objects", like `Moved IC ports`, `Resized IC`, `Placed Group`, `Create Circuit From Expression`, etc.

LeonMontealegre (approved):

PR #1403, opened 2025-05-16
Improved Transform/Rect usage and intersection testing

### LINKED ISSUE #1375: Model Rewrite: AABB hit tests

LeonMontealegre:
Also lots of assembler changes since Transform's now scale local-space to be a unit square, so pretty much every *number* had to be changed D: (it's a lot nicer now though)

Fixes #1375


src/projects/shared/api/circuit/src/utils/math/Transform.ts
```diff
@@ -1,100 +1,46 @@
...
- * For performance reasons the transform also stores a list of corners
- *  to be able to quickly apply intersection testing.
+ * A Transform holds all the spacial information about an object: position, rotation, and scale.
+ * A Transform can be thought of as a rectangle at the given pos, angle, and with size = scale.
+ *
+ * In local space of a Transform, the unit square represents the bounds of the rectangle for the Transform.
+ *
+ * For performance reasons the transform also stores a list of corners to be able to quickly apply intersection testing.
+ *
+ * It should be noted that the Vector is immutable and its values can only be set on construction.
```
TGCrystal, line 17:
```suggestion
 * It should be noted that the Transform is immutable and its values can only be set on construction.
```

PR #1462, opened 2025-06-29
Fix fit to screen when vertical is larger than horizontal

### LINKED ISSUE #1461: Model Rewrite: Fit to screen not fully working
Smart Place ASCII Display then press f. Final result is too zoomed in

Image

Selecting all components changes nothing

TGCrystal:
closes #1461

### DIFF
--- a/src/projects/shared/api/circuitdesigner/src/public/impl/Viewport.ts
+++ b/src/projects/shared/api/circuitdesigner/src/public/impl/Viewport.ts
@@ -341,11 +341,12 @@ export class ViewportImpl<T extends CircuitTypes> extends MultiObservable<Viewpo
         const marginSize = V(left + right, bottom + top);
         const bbox = Rect.Bounding(objs.map(({ bounds }) => bounds));
 
-        const screenSize = canvasInfo.screenSize.sub(V(left, bottom));
-        const worldSize = this.toWorldPos(screenSize).sub(this.toWorldPos(V(0, 0)));
+        const screenSize = canvasInfo.screenSize.sub(V(marginSize.x, marginSize.y));
+        // this.toWorldPos(screenSize) gets a negative number for y which is why we scale by -1, see #1461
+        const worldSize = this.toWorldPos(screenSize).sub(this.toWorldPos(V(0, 0))).scale(V(1, -1));
 
         // Determine which bbox dimension will limit zoom level
-        const ratio = V(bbox.width / worldSize.x, -bbox.height / worldSize.y);
+        const ratio = V(bbox.width / worldSize.x, bbox.height / worldSize.y);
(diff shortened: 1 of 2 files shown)


src/projects/shared/api/circuitdesigner/src/public/impl/Viewport.ts
```diff
@@ -336,16 +336,16 @@ export class ViewportImpl<T extends CircuitTypes> extends MultiObservable<Viewpo
...
-        const marginSize = V(left - right, bottom - top);
+        const marginSize = V(left + right, bottom + top);
         const bbox = Rect.Bounding(objs.map(({ bounds }) => bounds));
 
         const screenSize = canvasInfo.screenSize.sub(V(left, bottom));
         const worldSize = this.toWorldPos(screenSize).sub(this.toWorldPos(V(0, 0)));
 
         // Determine which bbox dimension will limit zoom level
-        const ratio = V(bbox.width / worldSize.x, bbox.height / worldSize.y);
+        const ratio = V(bbox.width / worldSize.x, -bbox.height / worldSize.y);
```
LeonMontealegre, line 348: Will the negative sign not always make it negative meaning the .x value will always be chosen in Math.max below?

src/projects/shared/site/src/containers/MainDesigner/index.tsx
```diff
@@ -46,6 +46,10 @@ export const MainDesigner = ({ otherPlace }: Props) => {
         dispatch(SetCircuitSaved(false))),
     [designer]);
 
+    useEffect(() => {
+        designer.viewport.margin = { top: HEADER_HEIGHT };
```
LeonMontealegre, line 50: Do this in the useLayoutEffect when setting `window.Circuit` or maybe just even in the root `index.tsx` when creating the designer?

LeonMontealegre (changes requested):

src/projects/shared/api/circuitdesigner/src/public/impl/Viewport.ts
TGCrystal: The issue was that this was always computing to be negative because worldSize.y is always negative because `this.toWorldPos(screenSize)` results in a negative value for y (and I don't really follow the math any further, I'm just copying from master and master has this same issue). I can put the `-` on the worldSize.y so it is more clear what it is negating

src/projects/shared/site/src/containers/MainDesigner/index.tsx
TGCrystal: Unless I missed it, index.tsx is in digital and I think this logic could be in shared. I'll put it in the other useLayoutEffect then

PR #1502, opened 2025-07-12
Miscellaneous eslint fixes

TGCrystal:
Just some cleanup before I attempt the eslint v9 upgrade. Doesn't fix all errors, most errors left are either in analog or are react-compiler linting rule fails.


src/projects/shared/site/src/containers/Header/Right/SettingsMenu/index.tsx
```diff
@@ -30,7 +30,7 @@ export const SettingsMenu = () => {
     //  especially the gross 'useAPIMethods' thing
     useEffect(() => {
         designer.viewport.debugOptions = debugInfo;
-    }, [designer, ...Object.values(debugInfo)]); // Updates when any of the debugInfo values change
+    }, [debugInfo, designer]); // Updates when any of the debugInfo values change
```
LeonMontealegre, line 33: does this work though? these dependency lints actually changed behavior so I'm not as keen on letting them through like that

LeonMontealegre (approved):

PR #1506, opened 2025-08-17
Some linting fixes

TGCrystal:
Some easier linting fixes. Also fixes the ignores so it doesn't pick up .js files, cuts the number of errors down a lot. Now down to:
image

And I did check that loading still works after the reader change there


src/projects/digital/site/src/containers/SelectionPopup/modules/CreateICButtonModule.tsx
```diff
@@ -33,10 +34,10 @@ export const CreateICButtonModule = ({ circuit }: Props) => {
         return;
 
     const open = () => {
-        if (icValidationStatus !== ICValidationStatus.Valid)
-            setShowError(true);
+        if (icValidationStatus === ICValidationStatus.Valid)
+            {dispatch(OpenICDesigner(props["ids"]));}
```
LeonMontealegre: nit: it added brackets around the statement

src/projects/shared/api/circuit/tests/public/Component.test.ts
```diff
@@ -251,7 +249,7 @@ describe("Component", () => {
     });
 
     describe("isNode", () => {
-        test("Test component is not a node", () => {
+        test("Component test component is not a node", () => {
```
LeonMontealegre: This doesn't read correctly in my brain, what's the linting error here? Test name shouldn't start with "test"? Should it just be like `Check component is not a node`?

src/projects/shared/site/src/utils/ImageExporter.ts
```diff
@@ -1,4 +1,4 @@
-import jsPDF from "jspdf";
+import { jsPDF } from "jspdf";
```
LeonMontealegre, line 1: Did you test that this works? These are different statements AFAIU?

src/projects/shared/site/src/utils/Importer.ts
```diff
@@ -1,17 +1,3 @@
...
-            resolve(reader.result);
-        });
-        reader.addEventListener("abort", () => reject("Failed to load file!"));
-        reader.addEventListener("error", () => reject("Failed to load file!"));
-        if (type === "string")
-            reader.readAsText(file);
-        else
-            reader.readAsArrayBuffer(file);
-    });
+    return type === "string" ? file.text() : file.arrayBuffer()
```
LeonMontealegre, line 2: Did you test that this works?

src/projects/shared/site/src/utils/LoadingScreen.ts
```diff
@@ -30,7 +30,8 @@ export async function LoadingScreen(
 
     let errored = false;
     let curPercent: number;
-    setProgress((curPercent = initialPercent));
+    curPercent = initialPercent;
```
LeonMontealegre: nit: just initialize it when declaring it in the line above

src/projects/shared/api/circuit/tests/public/Component.test.ts
TGCrystal: I have absolutely no memory of why this changed so I'll change it to `Check component is not a node`

src/projects/shared/site/src/utils/ImageExporter.ts
TGCrystal: The jspdf module exports jsPDF as default so for this instance, it works the same. Tested just now

src/projects/shared/site/src/utils/Importer.ts
TGCrystal: Yes, I tested with some of the test circuits in the repo

LeonMontealegre (approved):

PR #1507, opened 2026-05-13
Begin replacing eslint with oxlint

### LINKED ISSUE #1473: Investigate replacing ESLint with Biome
Updating ESLint is gong to require changing the config file format to their new flat config. This will be a pain, and at that point it is worth investigating Biome as it is much faster and comes with support for linting css and formatting. There may be some rules we use in ESLint that are not present in Biome, so we will have to see what they are, if we want them, and if there is any value in running Biome for most rules and ESLint for the ones not yet in Biome.

TGCrystal:
Branched from #1506 

Total time to lint the whole project went from 71s to 3.4s

Still a few linting errors in there, mostly react hook rules and cyclical imports

Closes #1473


src/other/pages/docs/src/components/VariableInput/index.tsx
```diff
@@ -36,8 +37,8 @@ export const useVariable = (id: string) => {
 
 
 export interface InputFieldProps {
-    id: string;
-    defaultValue?: string;
+    readonly id: string;
```
LeonMontealegre, line 40: Are these a lot of the same changes in your other PR? Maybe the other one should be merged first?

package.json
```diff
@@ -68,8 +66,7 @@
     "eslint-plugin-opencircuits": "file:./scripts/linting",
     "eslint-plugin-playwright": "^2.2.0",
     "eslint-plugin-react": "^7.37.5",
-    "eslint-plugin-react-compiler": "^19.1.0-rc.2",
```
LeonMontealegre, line 71: Why remove react compiler?

TGCrystal: This is just the eslint rules for it, which are now part of eslint-plugin-react-hooks (see https://react.dev/blog/2025/10/07/react-compiler-1#migrating-from-eslint-plugin-react-compiler-to-eslint-plugin-react-hooks)

LeonMontealegre (dismissed):

LeonMontealegre (approved):

PR #1509, opened 2026-07-03
Add a ChangeLog popup & Version Info to the site

### LINKED ISSUE #982: Changelog Popup
It would probably be confusing for an OpenCircuits user who isn't up-to-date with RCOS or the Github to suddenly have the UI and other things change when logging onto the site.

It would be good to have a changelog popup appear every time a new OpenCirucits version is released, that has a list of all the changes made from the update.

**Additional context**
This would require some cookie to keep track of which version the user has last used, and then when that version != the current version, the popup should be displayed. Similarly, we should maybe have a button to show the changelog somewhere on the SideNav or maybe the Help Menu.

It should probably **not** show the popup when the user first opens the site for the first time, i.e., if they do not have this cookie storing the version. We want to show this popup only to users who have used the site before.

### LINKED ISSUE #1505: Add version information to the site
An indicator of the running version should be added to the site somewhere.  This should at least be the release version for production-type deployments.  Maybe this could be based on git tag or HEAD commit and injected in the build system.

Attached photo as a suggestion for where to put it: 
Image

LeonMontealegre:
Fixes #982
Fixes #1505

AI lowkey cooked with the popup design:
image

The version info hyperlinks to the commit of the build, but displays as the manually-set-major-version
image

Also added an initial .agents/AGENTS.md file


.agents/AGENTS.md
```diff
@@ -0,0 +1,21 @@
...
+- **`src/server/`**: The backend server written in Go.
+- **`src/other/pages/`**: The React frontends for the `landing` site and `docs` site.
+- **`docs/`**: The actual markdown documentation content for the docs site.
+- **`scripts/`**: Custom Node.js scripts for building, testing, and Dockerfiles for WASM compilation.
+- **Local Documentation**: `README.md` files exist throughout the directory tree. Always read them for vital context on local architecture and boundaries before making changes.
+
+## Core Commands
+- **Building**: `yarn build` (interactive) or `yarn build:prod src/projects/digital/site` etc.
+- **Testing**: `yarn test --ci <path>` (e.g., `yarn test --ci src/projects/analog/api/circuit`). Always use the `--ci` flag for agent environments.
+- **Linting**: `yarn lint:all`.
```
TGCrystal: lint:all doesn't exist anymore with oxlint (it only was there because eslint was so slow I set up that whole system to lint only the diff)

TGCrystal (dismissed):

